PR #4, opened 2019-02-27
Upgrade gdal version to 2.4

Alanamosse:
Fix a misprint that set gdal to the version 2.1 instead of 2.4 when the names specified 2.4.
A few test files are updated to match the new version outputs.
geos version is also updated to the latest version 3.7.1

Sorry about the mistake in the first place :)


tools/gdal/gdal_macros.xml
```diff
@@ -2,8 +2,8 @@
     <token name="@VERSION@">2.4.0</token>  
     <xml name="gdal_requirements">
         <requirements>
-            <requirement type="package" version="2.1.0">gdal</requirement>
-            <requirement type="package" version="3.5.0">geos</requirement>
+            <requirement type="package" version="2.4.0">gdal</requirement>
```
bgruening: shouldn't this be @VERSION@?

Alanamosse: Yes ofc 
``<requirement type="package" version="@VERSION@">gdal</requirement>``

Alanamosse:
"Funny" how it fails now but not the previous time

PR #7, opened 2019-03-20
Add Rdata manipulation tools

Alanamosse:
Two new and small tools, useful to manipulate rdata files very common in the ecological community

1- Load and give insight on a rdata file.
2- Extract the wanted attributes

I'm almost certain that the Ecology TS category is not optimal. 
Maybe a more data oriented category would be better : "converts format" or "Data source" ?


yvanlebras:
Groovvy baby!  Yes, I think we have to put these tools into "Ecology" + "Convert formats" main TS categoies. "Data source" is not relevant I think....
One thing, if I well remember, we can't extract a workflow where these tools are used... If I well remember because of the `<options from_dataset="rdata_attributes"> <column name="value" index="0"/>` parameter no ?

Alanamosse:
Yes this option can be problematic for workflow. 
Also the parser tool 'discovers' created datasets because of the undefined number of outputs, so you can't really define an output you want to use in a workflow.

PR #11, opened 2019-06-05
Small fixes for vigiechiro tools

Alanamosse:
In .shed.yml :
 - Correction for the remote repository url.

In BilanEnrichiPF.R  : 
 - Add encoding to accept french special char. 
 - Correct a sep causing wrong display in output hourly table

### DIFF
--- a/tools/vigiechiro/BilanEnrichiPF.R
+++ b/tools/vigiechiro/BilanEnrichiPF.R
@@ -29,7 +29,7 @@ if(substr(IdC2$`nom du fichier`[1],2,2)!="a")
   print("Wrong protocol, please only use this tool for a \'Point Fixe\' protocol.")
 }else{
   refPF=fread(args[2],encoding="UTF-8")
-  GroupList=fread(args[3],encoding="Latin-1")
+  GroupList=fread(args[3],encoding="UTF-8")
   
   IdC2$ConfV[is.na(IdC2$ConfV)]=""
   
--- a/tools/vigiechiro/BilanEnrichiPF.R
+++ b/tools/vigiechiro/BilanEnrichiPF.R
@@ -21,15 +21,15 @@ f2p <- function(x) #get date-time data from recording file names
 
 
 
-IdC2=fread(args[1])
+IdC2=fread(args[1],encoding="UTF-8")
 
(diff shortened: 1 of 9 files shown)


tools/vigiechiro/BilanEnrichiPF.R
```diff
@@ -232,6 +232,6 @@ if(substr(IdC2$`nom du fichier`[1],2,2)!="a")
   
   saveWidget(SummHTMLH,"output-hourly.html")
 #  write.csv2(SummPartH,"output-hourly.tabular",row.names=F)
-  write.table(SummPartH,"output-hourly.tabular",row.names=F,sep="\n")
+  write.table(SummPartH,"output-hourly.tabular",row.names=F,sep="\t")
```
bgruening: Isn't this tested :)

tools/vigiechiro/BilanEnrichiPF.R
```diff
@@ -21,15 +21,15 @@ f2p <- function(x) #get date-time data from recording file names
...
 
 if(substr(IdC2$`nom du fichier`[1],2,2)!="a")
 {
 #  print("Protocole non conforme, ce script doit etre lance uniquement pour un protocole Point Fixe")
   print("Wrong protocol, please only use this tool for a \'Point Fixe\' protocol.")
 }else{
-  refPF=fread(args[2])
-  GroupList=fread(args[3])
+  refPF=fread(args[2],encoding="UTF-8")
+  GroupList=fread(args[3],encoding="Latin-1")
```
bgruening: Why is this Latin, UTF does not work? Mh this is wired.

yvanlebras: I agree... I think the problem is the original encoding of the file... We are making test on it

tools/vigiechiro/BilanEnrichiPF.R
Alanamosse: The test compare one out of 3 files, it's pretty cheap you're right :D 
I'll had more

tools/vigiechiro/BilanEnrichiPF.R
Alanamosse: Re-encoding a file resolved this, we should be able to use UTF-8
Thanks @yvanlebras

PR #15, opened 2019-06-14
Spocc fixes

Alanamosse:
The rspocc package (v.0.9.0) does not have the same databases available.
Comment the broken datasources for now, until a new version fix this.
(Not sure about it, but probably some API changes broke accesses).

Plus correct a typo vernet --> **vertnet**


Alanamosse:
Thank you Björn !

PR #22, opened 2020-05-09
STOC and netcdf improvements

yvanlebras:


bgruening:
@yvanlebras I think you need to bump the tool version.

yvanlebras:
Oh, thank you Björn! Nice catch! Done !!!!!

info_file.txt
```diff
@@ -0,0 +1,84 @@
+netcdf dataset_2 {
```
bgruening, line 1: @yvanlebras the two news files, for what are they good for? They are not used in this PR afaik.

yvanlebras: hmmm really sorry, don't know how these 2 files (test-data from the nectdf GH branch) are arrive here ... AGAIN.... TTHHHHAAAAANNK   YYOUUUUUU you and your eagle eyes ;)

PR #23, opened 2020-05-22
fix shd.yml

bgruening:


bgruening:
here are a few errors: https://travis-ci.org/github/galaxyecology/tools-ecology/jobs/689937215

annefou:
Ok. Yes. I did not realize we had to update .shd.yml!

bgruening:
The old one uses very specific syntax, that you need to update every time you add a tool. The suite syntax does most of the stuff automagically.

bgruening:
Thanks.

annefou:
@bgruening Ok. That's my fault (not a surprise...). xarray_select.xml is not correct for tests. I can make a PR to your branch xarray_shed.

annefou:
I did not comment in the rigth thread...

Can we add something about the support of dask in xarray (http://xarray.pydata.org/en/stable/dask.html) for the long description:


**xarray (formerly xray) is an open source project and Python package that makes working with
  labelled multi-dimensional arrays simple, efficient, and fun! xarray integrates with Dask to support parallel computations and streaming computation on datasets that don’t fit into memory.**

It does not apply for the simple tools we have deployed but the plan is to add some more tools.

annefou:
Thank you!

bgruening:
Ok. So let's see if this pops-up in the TS now :)

Thanks @annefou

bgruening:
Done, it uploaded.

PR #24, opened 2020-05-22
fix shed_lint problems for xarray_select

annefou:


annefou:
@bgruening I think now travis does not fail (https://travis-ci.org/github/galaxyecology/tools-ecology); if I looked at the right place...

bgruening:
Anything you want to change in the shed.yml?

annefou:
Maybe we could add something about the support of dask in xarray (http://xarray.pydata.org/en/stable/dask.html)

xarray integrates with Dask to support parallel computations and streaming computation on datasets that don’t fit into memory.

It does not apply for the simple tools we have deployed but the plan is to add some more tools.

PR #25, opened 2020-05-26
Add CSV output format

yvanlebras:
so you can combine this tool with xarray netCDF ones ;)

### DIFF
--- a/tools/gdal/ogr2ogr.xml
+++ b/tools/gdal/ogr2ogr.xml
@@ -38,6 +38,7 @@
     <inputs>
         <param type="data" name="input" format="shp,xml,json,tabular,txt" label="Gdal supported input file" help="http://www.gdal.org/ogr_formats.html"/>
         <param type="select" name="f" label="Conversion format" value="" help="Select the format to convert your input file. Currently supported : Shapefile, GeoJSON, GML, KML and WFS.">
+            <option value="CSV">CSV</option>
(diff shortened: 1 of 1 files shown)


bgruening:
Mh, should xarray maybe consume tsv?

yvanlebras:
xarray can already consume tsv/tabular/csv (as tabular) files with coordinates so you can use the CSV output of this tool applied to an input shapefile GIS format (for example, as this is the most used format to save in situ "collection" points) to extract values from climate (or oceanographic or others) dataset generally stored as netCDF file. Hope this is understandable ;) In fact, supporting this CSV output format for GDAL was in the to-od list, even if no xarray tool exist as it is really usefull for researcher to extract coordinates informations from shapefiles into a text based format.

bgruening:
Ok, I have no clue. But you need CVS and it does not make sense to convert this file to TSV before we output it?

I'm just asking because manipulating TSV is easier in Galaxy then CSV, also TSV can be displayed.

yvanlebras:
Yes, you're right about CSV manipulation in Galaxy. Here, gdal output only csv files, no tsv ones. And it appears that the csv is well formatted so I can display it and use it easily in Galaxy. So I prefer, for now, "just" keep the "original" output GDAL csv format. If I see there is issues regarding csv, one would consider to export tsv instead of csv. It's ok for you ?
THANK YOU !!!

bgruening:
Thanks!

PR #30, opened 2020-07-22
add EDAM topics

yvanlebras:


tools/data_manipulation/xarray/xarray_select.xml
```diff
@@ -261,7 +261,10 @@
              </output>
         </test>
     </tests>
-
+    <edam_topics>
```
bgruening, line 264: could you not do this in macro?

yvanlebras:
True.... Was hesitating but wanna not modyfying too much things there... But I think we can rethink this on a next edam / bio.tools round related to additioon of more tags ?

bgruening:
Sure! :)

yvanlebras:
Thank you!

PR #32, opened 2020-10-15
Modifications of GLM tools and add new plot tool

ColineRoyaux:
New options (Year as continuous in GLM, ...) in GLM tools and new tool to create time series ggplot from your GLM results

### DIFF
--- a/tools/PAMPA/PAMPA_PLOT_GLM.xml
+++ b/tools/PAMPA/PAMPA_PLOT_GLM.xml
@@ -27,7 +27,7 @@
        </collection>
     </outputs>
     <tests>
-        <test> 
+        <test expect_num_outputs="2"> 
             <param name="input" value="GLM_table_population_analysis_on_Presence_absence_sansszcl_cropped.tabular"/>
             <param name="datatable" value="Presence_absence_table_sansszcl_cropped.tabular"/>
             <param name="unitobstable" value="Unitobs.tabular"/>
--- a/tools/PAMPA/PAMPA_PLOT_GLM.xml
+++ b/tools/PAMPA/PAMPA_PLOT_GLM.xml
@@ -1,0 +1,143 @@
+<tool id="pampa_plotglm" name="Create a plot from GLM data" version="@VERSION@">
+    <description>as temporal trend</description>
+    <macros>
+        <import>pampa_macros.xml</import>
+    </macros>
+    <expand macro="Plot_requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+        Rscript 
+         '$__tool_directory__/FunctExePlotGLMGalaxy.r' 
+         '$input'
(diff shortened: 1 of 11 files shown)


tools/PAMPA/pampa_macros.xml
```diff
@@ -67,6 +69,11 @@
     <xml name="pampa_output_GLM">
         <data name="output_recap" from_work_dir="GLMSummaryFull.txt" format="txt" label="Simple statistics on chosen variables on ${on_string}"/>
         <data name="output_rate" from_work_dir="RatingGLM.txt" format="txt" label="Your analysis rating file on ${on_string}"/>
+        <collection type="list" name="output_GLM">
```
bgruening, line 72: can we add a test for this outpu?

tools/PAMPA/PAMPA_PLOT_GLM.xml
```diff
@@ -0,0 +1,143 @@
...
+        <param name="unitobstable" type="data" format="tabular" label="Unitobs table used for glm" help=""/>
+    </inputs>
+    <outputs>
+       <data name="report" from_work_dir="stdout.txt" label="Report" format="txt"/>
+       <collection type="list" name="plots">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+)\.png" visible="false" format="png"/>
+       </collection>
+    </outputs>
+    <tests>
+        <test>
```
bgruening, line 30: can you add `expected_num_outputs` (https://docs.galaxyproject.org/en/latest/dev/schema.html#id78) here and maybe also in other tests?

bgruening (commented):
Thanks @ColineRoyaux, this looks very nice! A few comments inside.

* we need to bump the tool versions of all tools
* @yvanlebras with this "special" shed.yml we probably need to update this file as well, correct?

tools/PAMPA/pampa_macros.xml
ColineRoyaux: It is an optional output hence I didn't used it for the test but I'll do it now ! Thank you !

tools/PAMPA/PAMPA_PLOT_GLM.xml
ColineRoyaux: Thank you very much ! I'll do that

yvanlebras:
Version bumped! (If I am not saying a mistake)
Tank you so much Björn !!!

bgruening (approved):

bgruening:
Wuhu! Cool beans!

yvanlebras:
Wonder Coline "on the place"!!!!

ColineRoyaux:
Thank you very much ! 😄

yvanlebras:
Thank you Björn ! THANK YOU Coline !!!!!

PR #33, opened 2020-10-26
Add DHARMa package properly and tiny corrections

ColineRoyaux:
DHARMa package source code deleted from PAMPA files and added to requirements 
Corrections on latest tool (create plot)


pampa_macros.xml
```diff
@@ -0,0 +1,89 @@
+<macros>
```
bgruening, line 1: this file seems to be wrong in this folder

yvanlebras: oh right! Thanks! Will delete it, but need to be sure this is not the version we need on the right folder...

PR #34, opened 2020-10-31
update tests

bgruening:
@yvanlebras something fails here. It was never pushed to the TS because Travis was outdated. I fixed Travis, but now those tools are failing :(


bgruening:
Travis is here: https://travis-ci.org/github/galaxyecology/tools-ecology

bgruening:
We should switch to github actions. I can give you a hand on this. You only need to set the credentials.

yvanlebras:
Thank you Björn for seeing this! Will look at it. Duno what means "switching to gh actions" but will look at it + to Travis errors with @ColineRoyaux . THANK YOU!

bgruening:
Github actions. This is a similar service to Travis, which we use in IUC and are happy with it. I just need 15min of your time and we can remove travis and setup Github Actions.

ColineRoyaux:
@bgruening thank you very much ! After checking it seems like updated test files haven't been uploaded properly onto GitHub I'm sure I tried to upload them but something must have gone wrong without me noticing sorry !

bgruening:
@ColineRoyaux there is also now a new method in Galaxy where you can test just for the filesize. So you don't need to have the RData in this repository.

ColineRoyaux:
@bgruening Thank you ! I did it and commited on my current PR https://github.com/galaxyecology/tools-ecology/pull/33

PR #42, opened 2021-05-17
Xarray mapplot

annefou:
To be able to plot 2D xarray on a geographical map (assuma lat/lon or latitude/longitude coordinates).

Usage:
```
python xarray_mapplot.py birch_pollen_fc_global.netcdf 'bpg_conc' --verbose --shift \
 --land=0.1  --ocean=0.1 --coastline=0.2 --borders=0.1 --cmap "twilight" \
 --label "Birch pollen grain (grains/m3)"  --title "Birch pollen forecast on 15 May 2021" \
 --threshold 100 --time="0,1" --xlim="-180,180"  --ylim="-90,90"
```

- [x] python script to plot netCDF on a map using xarray
- [x] Galaxy tool for using `xarray_mapplot.py`
- [x] Tests for testing Galaxy tool


bgruening:
Let me know when you need a review :)

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,204 @@
+<tool id="xarray_mapplot" name="NetCDF xarray map plotting" version="0.18.2">
```
bgruening, line 1: if you add profile="20.05" I think the HOME will be set to your CWD automatically. Always use a recent profile if you can.

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,204 @@
...
+        <requirement type="package" version="1.2">cmcrameri</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir output_dir &&
+    HOME=`pwd`  &&
+    mkdir -p `pwd`/mlp_tmpdir &&
+    MPLCONFIGDIR=`pwd`/mlp_tmpdir &&
+    echo "Galaxy xarray version @VERSION@" > output_dir/version.txt &&
+    python '$__tool_directory__/xarray_mapplot.py' '$input' '$var'
+    #if $time_idx!=""
```
bgruening, line 18: you should maybe cast this to str() before comapre.

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,204 @@
...
+        </param>
+        <param name="range" type="text" value="" label="Range of values for plotting e.g. minimum value abd maximum value (minval,maxval) (optional)" />
+        <param name="threshold" type="text" value="" label="Do not plot values below this threshold (optional)" />
+        <param name="borders" type="text" value="" label="Add country borders with alpha value [0-1] (optional)" />
+        <param name="coastline" type="text" value="" label="Add coastline with alpha value [0-1] (optional)" />
+        <param name="ocean" type="text" value="" label="Add ocean with alpha value [0-1] (optional)" />
+        <param name="land" type="text" value="" label="Add land with alpha value [0-1] (optional)" />
+        <param name="title" type="text" value="" label="Specify plot title (optional)" />
+        <param name="colorbar_label" type="text" value="" label="Set a label for colormap (optional)" />
+        <param name="cmap" type="text" value="" label="Specify which colormap to use for plotting (optional)" />
```
bgruening, line 100: do you know type="color" ?

https://docs.galaxyproject.org/en/latest/dev/schema.html#color

If you can not use it I think a better way is to have a long select box

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,204 @@
...
+        </param>
+        <param name="time_idx" type="text" value="" label="List of time indices such as 0,1,... (optional)" />
+        <param name="xlim" type="text" value="" label="longitudes values 'lonW,lonE' for limited geographical area (optional and only available with some projections)" />
+        <param name="ylim" type="text" value="" label="latitudes values 'latS,latN' for limited geographical area (optional and only available with some projections)" />
+        <param name="shift" type="select" display="radio" label="Shift longitudes [0,360] --> [-180,180]"> 
+             <option value="" selected="true">No</option>
+             <option value="Yes">Yes</option>
+        </param>
+        <param name="range" type="text" value="" label="Range of values for plotting e.g. minimum value abd maximum value (minval,maxval) (optional)" />
+        <param name="threshold" type="text" value="" label="Do not plot values below this threshold (optional)" />
```
bgruening, line 93: do all of them as optional="true"

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,204 @@
...
+         --label="$colorbar_label"
+    #end if
+    #if $title!=""
+         --title=$title
+    #end if
+    #if $cmap!=""
+         --cmap="$cmap"
+    #end if
+    #if $proj!=""
+         --proj="$proj"
```
bgruening, line 33: all single quote

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,204 @@
...
+        </collection>
+    </outputs>
+    <tests>
+        <test>
+             <param name="input" value="dataset-ibi-reanalysis-bio-005-003-monthly-regulargrid_1510914389133.nc"/>
+             <param name="var" value="phy"/>
+             <param name="var_tab" value="var_tab_dataset-ibi"/>
+             <conditional name="condi_source_coord">
+                 <param name="coord_source" value="coord_from_stdin"/>
+                  <conditional name="condi_coord">
```
bgruening, line 116: indentation ...

tools/data_manipulation/xarray/xarray_mapplot.xml
annefou: I think I find a better way to select times (from the list of coordinates).

tools/data_manipulation/xarray/xarray_mapplot.xml
annefou: No but here we need a colormap (such as matplotlib colormap or cmcrameri)

tools/data_manipulation/xarray/xarray_mapplot.xml
annefou: Great! Thank you! I always forget about this option!

tools/data_manipulation/xarray/xarray_mapplot.py
```diff
@@ -124,147 +124,11 @@ def __init__(self, input, proj, varname, cmap, output, verbose=False,
...
-        if self.cmap == "bamO_r":
-            self.cmap = cm.bamO_r
-        if self.cmap == "brocO_r":
-            self.cmap = cm.brocO_r
-        if self.cmap == "corkO_r":
-            self.cmap = cm.corkO_r
-        if self.cmap == "vikO_r":
-            self.cmap = cm.vikO_r
+    def get_cmap(self, cmap):
+        if cmap[0:3]=='cm.':
```
bgruening:
```suggestion
        if cmap.startswith('cm.'):
```

tools/data_manipulation/xarray/xarray_coords_info.xml
```diff
@@ -0,0 +1,53 @@
...
+    </inputs>
+    <outputs>
+        <collection type="list" name="output_dir" label="Coordinates">
+            <discover_datasets pattern="__name_and_ext__" visible="true" directory="output_dir"/>
+        </collection>
+    </outputs>
+    <tests>
+        <test>
+             <param name="input" value="dataset-ibi-reanalysis-bio-005-003-monthly-regulargrid_1510914389133.nc"/>
+              <output_collection name="output_dir">
```
bgruening, line 26:
```suggestion
              <output_collection name="output_dir" type="list" count="5">
```

bgruening: How big can such a collection be? One file per coordinate sounds dangerous to me :) Can this be 100.000 elements?

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,331 @@
+<tool id="xarray_mapplot" name="NetCDF xarray map plotting" profile="20.05" version="0.18.2">
+    <description>Visualize netCDF variables on a geographical map</description>
```
bgruening, line 2: You could include something like that:

```
    <macros>
        <token name="@TOOL_VERSION@">0.18.2</token>
        <token name="@VERSION_SUFFIX@">0</token>
    </macros>
```

This would also allow our new bot to bump the tool version automatically when there is a new xarray release.

tools/data_manipulation/xarray/xarray_coords_info.xml
annefou: no. it is related to the number of dimensions (latitude, longitude, depth, time, height, levels, etc.). So not so many (<20)

tools/data_manipulation/xarray/xarray_select.xml
```diff
@@ -65,6 +77,7 @@
             <param name="coord_source" type="select" label="Source of coordinates">
                 <option value="coord_from_stdin">Manually enter coordinates</option>
                 <option value="coord_from_file">Use coordinates from input file</option>
+                <option value="tolerance">Maximum distance between original and selected value for inexact matches e.g. abs(index[indexer] - target) <= tolerance</option>
```
bgruening, line 80:
```suggestion
                <option value="tolerance"><![CDATA[Maximum distance between original and selected value for inexact matches e.g. abs(index[indexer] - target) <= tolerance]]></option>
```

bgruening: This will hopefully fix the linting error that we see here.

tools/data_manipulation/xarray/macros.xml
```diff
@@ -0,0 +1,4 @@
+<macros>
```
bgruening, line 1: you actually can put this into the same XML tool file.
https://github.com/galaxyproject/tools-iuc/blob/5fe46c36adac4843cd548802073aa62b0afc61cd/tools/bellerophon/bellerophon.xml#L3

bgruening: If you like, but no wait you are correct that should be shared across the other tools.

annefou: I like `+galaxy@VERSION_SUFFIX@` (I did not really know what to do with `VERSION_SUFFIX`). Thanks!

tools/data_manipulation/xarray/test-data/version.tabular
```diff
@@ -0,0 +1 @@
+Galaxy xarray version 0.18.2
```
bgruening, line 1: one o those file is not needed?

annefou:
@bgruening I think I am done with this tool (for this version...). No need to make your review over the weekend; I can wait!. Thanks.

tools/data_manipulation/xarray/xarray_coords_info.xml
```diff
@@ -0,0 +1,56 @@
...
+    <outputs>
+        <collection type="list" name="output_dir" label="Coordinates">
+            <discover_datasets pattern="__name_and_ext__" visible="true" directory="output_dir"/>
+        </collection>
+    </outputs>
+    <tests>
+        <test>
+             <param name="input" value="dataset-ibi-reanalysis-bio-005-003-monthly-regulargrid_1510914389133.nc"/>
+              <output_collection name="output_dir" type="list" count="5">
+                <element name="time" file="time.tabular" />
```
bgruening, line 30: maybe you can add here `ftype="tabular"` this will test that the file is of the expected format.

tools/data_manipulation/xarray/xarray_coords_info.xml
```diff
@@ -0,0 +1,56 @@
...
+corresponding coordinate. The output of this tool is usually used as input to other tools.
+
+**Input**
+
+A netcdf file (xxx.nc).
+
+--------------------------------
+
+    ]]></help>
+
```
bgruening, line 55: for some reason lint is not running on CI ... every tool should have a citations tag, even if there is no citation

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,346 @@
...
+        </param>
+        <param name="range" type="text" optional="true" label="Range of values for plotting e.g. minimum value abd maximum value (minval,maxval) (optional)" />
+        <param name="threshold" type="float"  optional="true" label="Do not plot values below this threshold (optional)" />
+        <param name="borders" type="float" optional="true" label="Add country borders with alpha value [0-1] (optional)" />
+        <param name="coastline" type="float" optional="true" label="Add coastline with alpha value [0-1] (optional)" />
+        <param name="ocean" type="float" optional="true" label="Add ocean with alpha value [0-1] (optional)" />
+        <param name="land" type="float" optional="true" label="Add land with alpha value [0-1] (optional)" />
+        <param name="title" type="text" optional="true" label="Specify plot title (optional)" />
+        <param name="colorbar_label" type="text"  optional="true" label="Set a label for colormap (optional)" />
+        <param name="cmap" type="select" optional="true" label="Specify which colormap to use for plotting (optional)">
```
bgruening, line 117: those things should probably be moved at some point into the macro for reuse.

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,346 @@
...
+            <option value="spring">spring</option>
+            <option value="summer">summer</option>
+            <option value="tab10">tab10</option>
+            <option value="tab20">tab20</option>
+            <option value="tab20b">tab20b</option>
+            <option value="tab20c">tab20c</option>
+            <option value="terrain">terrain</option>
+            <option value="winter">winter</option>
+        </param>
+        <param name="proj" type="text" optional="true" label='Specify the projection (proj4) on which we draw e.g. {"proj":"PlateCarree"} with double quote (optional)'>
```
bgruening, line 267: why not only ask to type "PlateCarree"?

bgruening: Sorry for my naive question.

tools/data_manipulation/xarray/xarray_mapplot.xml
```diff
@@ -0,0 +1,346 @@
...
+
+This tool plot a variable on a geographical map. It must be a 2D variable (latitude, longitude) and eventually with 
+and additional time dimension (specific time to plot should then be selected).
+
+The output is a collection of plots (png format).
+
+-------------------------------------------------
+
+The xarray select tool can be used after the xarray Info and xarray coord.
+    ]]></help>
```
bgruening, line 345: citation?

bgruening: maybe as macro

tools/data_manipulation/xarray/xarray_tool.py
```diff
@@ -213,6 +244,14 @@ def selection_from_coords(self):
                             str(row.Index) + '.tabular')
             self.selection()
 
+    def get_coords_info(self):
+        ds = xr.open_dataset(self.infile)
+        for c in ds.coords:
+            filename = self.coords_info + '/' + c.strip() + '.tabular'
```
bgruening, line 250: please use os.path.join() here

tools/data_manipulation/xarray/test-data/version.tabular
annefou: Why? Should I always have `version.tabular`?

bgruening: Not always but you have two tools that generate two different files with the same content, is that like it should be?

btw. do you know: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-version-command

tools/data_manipulation/xarray/xarray_mapplot.xml
annefou: because in the Nordics, we use more complex projection and we need to specify addition parameters. for instance:
`{"proj":"EquidistantConic", "central_longitude": 20.0, "central_latitude": 70.0 }`

So I chose to use the standard proj4 "way" to specify projection.

bgruening: I see, maybe that example should be added to the help?

tools/data_manipulation/xarray/xarray_netcdf2netcdf.xml
```diff
@@ -0,0 +1,192 @@
...
+                        <options from_dataset="dim_tab">
+                            <column name="name" index="1"/>
+                            <column name="value" index="0"/>
+                        </options>
+                    </param>
+                </when>
+            </conditional>
+        </repeat>
+        <param type="float" name="scale" optional="true" help="scale factor (float) to apply to selection (optional)"/>
+        <param name="write_all" type="select"  optional="true" display="radio" label="Write all to new netCDF file or selection only">
```
bgruening, line 98: Is that not a classical Boolean?

tools/data_manipulation/xarray/xarray_netcdf2netcdf.xml
```diff
@@ -0,0 +1,192 @@
...
+                <when value="from">
+                    <param name="value" type="select" multiple="false" label="Choose the start value">
+                        <options from_dataset="dim_tab">
+                            <column name="name" index="1"/>
+                            <column name="value" index="0"/>
+                        </options>
+                    </param>
+                </when>
+            </conditional>
+        </repeat>
```
bgruening, line 96: Min/max?

bgruening (approved):

yvanlebras:
Amazing Anne, wunderschön Björn!!!! THANK you so much. Looking forward to test new version of the select tool! This can be a huge huge tool for ecology communities!!!

yvanlebras:
Thank you so much! Testing the new xarray selection tool version today make me see that this update fix errors I had with some particular dataset and/or operation! It seems to me there is maybe an error in the outputs as a "version" dataset has been added so we can manage collection output in cas of error (to help rerun the tool or just look at job information notably) but this dataset is not aside of the collection as another job output but inside the collection. Am I ok @annefou ? Moreover, this can be maybe more interesting to have stdout in this file instead of tool version ? Finally, not sure it's easy to know what kind of value can be written in the "Maximum distance between original and selected value for inexact matches e.g. abs(original, index) is lower or equal to tolerance" parameter. Maybe this is "just" due to the fact I never used this library , but maybe the explanation can be detailled in the help section or other?

Thank you so much Anne!!!! This is really a tool particularly usefull for ecology communities!!!

annefou:
Great. I think it is very valuable feedback. Should we make separate issues (so we can fix them for instance during the co-fest at GCC or during a papercut)?

yvanlebras:
Very good idea! Will try to create the issues today! Have a nice day Anne!

PR #43, opened 2021-05-17
fix flake8 errors for xarray_tool

annefou:
it looks like there is an error with flake8 for `xarray_tool.py` so better to fix it in a separate PR than in my branch (where I implement map plot).


bgruening:
Does that happen only locally or also here on github? We ignore usually the 80chars linting rule.

annefou:
It also appears on github and actually this is how I spot it.

annefou:
But happy to close this PR if it does not matter. The file has not been changed and it used to pass so it probably does not matter.

bgruening:
I will update the CI: https://github.com/galaxyecology/tools-ecology/pull/44

and take care, thanks!

PR #53, opened 2021-11-13
Add a first version of MetaShARK Interactive tool

yvanlebras:
Here is a first version of MetaShARK interactive tool GxIT. Here nothing is managed as input or output, need to work on it as there is no link between datasets from Galaxy or output.

### DIFF
--- a/tools/interactive/interactivetool_metashark.xml
+++ b/tools/interactive/interactivetool_metashark.xml
@@ -1,10 +1,13 @@
-<tool id="interactive_tool_metashark" tool_type="interactive" name="metashark" version="1.7">
+<tool id="interactive_tool_metashark" tool_type="interactive" name="MetaShARK" version="@TOOL_VERSION@">
     <description>Metadata Shiny Automated Resource and Knowledge</description>
+    <macros>
+        <token name="@TOOL_VERSION@">1.7.0</token>
+    </macros>
     <requirements>
-        <container type="docker">eliearnaud/metashark:1.7.0</container>
+        <container type="docker">eliearnaud/metashark:@TOOL_VERSION@</container>
     </requirements>
     <entry_points>
-        <entry_point name="metashark visualisation" requires_domain="True">
+        <entry_point name="MetaShARK visualisation" requires_domain="True">
             <port>3838</port>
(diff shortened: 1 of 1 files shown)


bgruening:
For the time beingz those ITs should go directly into the Galaxy main repository.

yvanlebras:
For now a version with no direct interaction with Galaxy (no input from Galaxy history and no output to Galaxy history)

yvanlebras:
What do you think @bgruening to propose to have such interactive tool where Galaxy can be seen "just" as a R Shiny apps ""scheduler"" ?

tools/interactive/interactivetool_metashark.xml
```diff
@@ -0,0 +1,55 @@
...
+    <command><![CDATA[
+
+        mkdir -p /srv/shiny-server/data/ &&
+        cp '$infile' /srv/shiny-server/data/inputdata.txt &&
+        R -e "options('shiny.port'=3838,shiny.host='0.0.0.0');MetaShARK::runMetashark()"
+
+    ]]>
+    </command>
+    <inputs>
+        <param name="infile" type="data" format="tabular,csv" label="csv file"/>
```
bgruening, line 21: fix label please

tools/interactive/interactivetool_metashark.xml
```diff
@@ -0,0 +1,55 @@
+<tool id="interactive_tool_metashark" tool_type="interactive" name="metashark" version="1.6">
```
bgruening, line 1: version number does not fit the container ? maybe as a macro token?

bgruening (commented):
@yvanlebras this should be submitted to the main Galaxy repo. ITs should not be put into the TS (yet).

tools/interactive/interactivetool_metashark.xml
yvanlebras: You're right! Thank you!

tools/interactive/interactivetool_metashark.xml
yvanlebras: Thank you !

yvanlebras:
Done!

PR #58, opened 2022-07-05
Timeseries tool

Quickbeasts51429:
Added time-series extraction tool.


annefou:
@bgruening 
> @Quickbeasts51429 very nice. Can you please add a shed.yml file?

We put everything we xarray tools. Also updating all xarray tools with the latest Python packages.

bgruening:
Tests seem to fail with: `Output collection 'output_dir': expected to have 5 elements, but it had 0.`

annefou:
> Tests seem to fail with: `Output collection 'output_dir': expected to have 5 elements, but it had 0.`

Ok. I will look at it. I probably messed up somewhere! Thanks a lot for pointing the error.

annefou:
> Tests seem to fail with: `Output collection 'output_dir': expected to have 5 elements, but it had 0.`

@bgruening I think I finally understood what was wrong! I was trying to use a version of geopandas from conda-forge that is in `cf-staging` (for some reasons most of the geopandas versions are in cf-staging mode).

bgruening:
Oh, that is unfortunate ... should we create a container?

annefou:
> Oh, that is unfortunate ... should we create a container?

What do you mean? Sorry, I don't get it!

annefou:
> If you really need the new version we can create a container and not use condor.

No we don't need a new version. It is not important. Thanks.

bgruening (approved):
LGTM!

PR #63, opened 2022-11-23
Champs blocs

Marie59:
Adding another tool to the champs blocs workflow


tools/champ_blocs/div_index.xml
```diff
@@ -0,0 +1,84 @@
...
+        Rscript
+            '$__tool_directory__/div_index.r'
+            '$input_data'
+            '$output_div'
+            '$output_rdata'
+            '$plots'
+        ]]>
+    </command>
+    <inputs>
+        <param name="input_data" type="data" format="data" label="Input Clean data"/>
```
bgruening, line 26: the type is is very generic, anything better that we can use here?

Marie59: I tried to put the format as Rdata but it does not detect the output 'Clean Rdata' from the cb_dissimlarity tool.

Marie59:
I modified the name of the tool dissimilarity (even though this tool was already merged before). Was this the way to change the name on Galaxy ?

Marie59:
OOOH I don't know why but it works !! Youhou !

tools/champ_blocs/div_index.xml
bgruening: Try `rdata` the real datatype is called that way see here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample#L670

Please chance it everywhere, I hope it will work then

Marie59:
Everything seems to be working with the format rdata !

bgruening (approved):

bgruening:
Cool!

PR #67, opened 2023-04-07
Srs update

Marie59:
Fix for the bil, hdr issue in pca_raster.xml and spectral_indices.xml.

### DIFF
--- a/tools/srs_tools/indices_spectral.xml
+++ b/tools/srs_tools/indices_spectral.xml
@@ -119,7 +119,9 @@
             <discover_datasets pattern="(?P&lt;designation&gt;.+)\.png" visible="false" format="png"/>
         </collection>
         <collection type="list" name="output_raster" label="${input_indice} raster">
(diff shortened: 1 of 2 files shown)


bgruening (approved):

yvanlebras:
The error on the preprocess tool seems to be "only" related to lack of memory. I think this is ok , moreover, There is no change on this tool if I am not saying a mistake

yvanlebras:
Thank you so much Marie !!!
Is it ok for you to merge Björn ?
Good appetit!

bgruening:
What is the test error?

yvanlebras:
exit code 137 => issue with container resources

bgruening:
I can merge and push locally later at the day. But we need to think about stability of those tools :-(

yvanlebras:
Yes. This tool is using a lot of memory, and it is very hard to find small data for the test, but we definitely need to find a way having these test using less resources. On the GH actions, we can't increase resources and/or this is not so easy ?

bgruening:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

yvanlebras:
I see that on a 35 Mb, preprocess tool is using 41 Gb memory...

PR #74, opened 2023-07-21
Add tools emlassemblyline

TanguyGen:


tools/EMLassemblyline/eml2eal.xml
```diff
@@ -21,12 +21,12 @@
...
             <output_collection name="EML_templates" type="list">
-				<element file="annotations.txt" ftype="tabular"/>
-				<element file="keywords.txt" ftype="tabular" />
-				<element file="attributes_data_blary_&amp;_al.txt" ftype="tabular"/>
-				<element file="catvars_data_blary_&amp;_al.txt" ftype="tabular"/>
-				<element file="geographic_coverage.txt" ftype="tabular" />
-				<element file="personnel.tabular" ftype="tabular" />
+				<element name="annotations" file="annotations.txt" ftype="tabular"/>
+				<element name="keywords" file="keywords.txt" ftype="tabular" />
+				<element name="attributes_data_blary_&amp;_al" file="attributes_data_blary_&amp;_al.txt" ftype="tabular"/>
```
bgruening: You are controlling the R code so maybe we can avoid the & and other strange characters.

tools/EMLassemblyline/make_eml.xml
```diff
@@ -0,0 +1,67 @@
...
+    <description> Create EML from EAL templates</description>
+    <requirements>
+        <requirement type="package" version="4.3.1">r-base</requirement>
+        <requirement type="package" version="3.5.5">r-emlassemblyline</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+    mkdir output_template &&
+    cd output_template &&
+    #for $input in $templates
+        cp '$input' '${input.element_identifier}.txt' &&
```
bgruening, line 11: why cp? does ln work?

tools/EMLassemblyline/make_eml.xml
```diff
@@ -0,0 +1,67 @@
...
+    #end for
+    find . -type f -empty -print -delete &&
```
bgruening, line 13: why is this needed? Can you not check that inside the R script?

tools/EMLassemblyline/make_eml.xml
```diff
@@ -0,0 +1,67 @@
...
+    Rscript '$__tool_directory__/make_eml.R'
+    ]]></command>
+    <inputs>
+        <param name="templates" type="data_collection" type_collection="list" format="tabular" label="Upload all EAL templates" help= "Input a data collection." />
```
bgruening, line 17: Why are data collection needed. Sorry I missed the discussion.

Usually multiple="true" is enough?

tools/EMLassemblyline/make_eml.xml
```diff
@@ -0,0 +1,67 @@
...
+    </inputs>
+    <outputs>
+        <data name="metadata" format="xml" label="Metadata">
+            <discover_datasets directory = "output_template" pattern= "(?P&lt;designation&gt;.+)\.xml" visible="true"/>
+        </data>
+    </outputs>
+    <tests>
+        <test expect_num_outputs="1"> 
+            <param type="data_collection" name="templates" value="Assessing_the_importance_of_field_margins_for_bat_species.xml">
+                <collection type="list">
```
bgruening, line 27: Is it __always__ those 7 files? Or can be more, or less?

tools/EMLassemblyline/make_eml.xml
```diff
@@ -0,0 +1,67 @@
...
+                    <element file="annotations.tabular"/>
+                    <element file="keywords.tabular" />
+                    <element file="attributes_data_blary_al.tabular"/>
+                    <element file="catvars_data_blary_al.tabular"/>
+                    <element file="keywords.tabular" />
+                    <element file="geographic_coverage.tabular" />
+                    <element file="personnel.tabular" />
+                </collection>
+            </param>
+            <output name="metadata" ftype="xml" file="Metadata.xml"/>
```
bgruening, line 37: this test expects a single output, but you have defined in the output section a collection as output.

tools/EMLassemblyline/eml2eal.xml
```diff
@@ -0,0 +1,53 @@
+<tool id="eml2eal" name="EML to EAL" version="0.1.0+galaxy0">
```
bgruening, line 1:
```suggestion
<tool id="eml2eal" name="EML to EAL" version="0.1.0+galaxy0" profile="22.05">

```

tools/EMLassemblyline/eml2eal.xml
```diff
@@ -0,0 +1,53 @@
...
+		<requirement type="package" version="4.3.1">r-base</requirement>
+	   <requirement type="package" version="3.5.5">r-emlassemblyline</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+       Rscript
+         '$__tool_directory__/eml2eal.R'
+         '$data'
+    ]]></command>
+    <inputs>
+		<param name="data" type="data" format="xml" label="Upload EML" help="Insert only metadata using EML standard"/>
```
bgruening, line 13: maybe a better name than data?

tools/EMLassemblyline/.shed.yml
```diff
@@ -0,0 +1,44 @@
...
+    - Ecology
+description: Tools to operate machine actionnable jobs to EML metadata
+long_description: |
+  Tools to operate machine actionnable jobs to EML metadata from MetaShRIMPS
+name: metashrimps 
+owner: ecology 
+remote_repository_url: https://github.com/tanguygen/metacure
+type: unrestricted
+
+repositories:
```
bgruening, line 11: please remove this and adopt this shed.yml to a suite based yml file: https://github.com/galaxyproject/tools-iuc/blob/acc495bd77aa3defdda7ec8d8b84a37d6832f2ee/tools/dram/.shed.yml#L17

tools/EMLassemblyline/eml2eal.xml
yvanlebras: The names of the outputs files are coming from the metadata file describing the data files... We can put rules on the R code to try look at all strange characters but it seems to me potentially better to let the original choice... You think we can't manae having & character on names here?

yvanlebras: Finally I modified the input XML metadata file to 'ot have these & characters

tools/EMLassemblyline/make_eml.xml
yvanlebras: Yes, multiple = true can be ok, but here, the idea is to create several tools who can deal with these "metadata templating files" who are supposed to be in a same folder. Using data collection is a way to have a similar manner to work on these templating files gathered in the same "place"

tools/EMLassemblyline/make_eml.xml
yvanlebras: It is depending the content of the input file

tools/EMLassemblyline/eml2eal.xml
yvanlebras: Done, THANK YOU Björn!

tools/EMLassemblyline/make_eml.xml
yvanlebras: Seems ok no?

bgruening: output collection are specified this way?

https://github.com/galaxyecology/tools-ecology/pull/74/files#diff-7e85ffa1db3b0bc3de3c0649530c0e07f5fd1deaa19f24d2dc2a8f48bc702f36R24

Is the output that you expect from this tool a collection or a single dataset?

yvanlebras: A single dataset. This tool create a metadata file following EML metadata standards specification from existing templates metadata files.

bgruening: Than the output section needs to be adopted

tools/EMLassemblyline/.shed.yml
yvanlebras: Done! THANK YOU!!!

tools/EMLassemblyline/make_eml.xml
yvanlebras: ln -s is working properly! THANK YOU WONDER Björn!!!!

tools/EMLassemblyline/make_eml.xml
yvanlebras: Not so important I think. I remove it and if needed, we can manage this differently

tools/EMLassemblyline/make_eml.xml
```diff
@@ -34,7 +34,7 @@
             </param>
             <output name="metadataout" ftype="xml" count="1" >
                 <assert_contents>
-					<has_line line="<?xml version="1.0" encoding="UTF-8"?>" />
+					<has_size value="6700" delta="200" />
```
bgruening, line 38: Checking for the file content is usually better than checking for the size.
For XML in particular you have those nice XML assertions: https://docs.galaxyproject.org/en/latest/dev/schema.html#assertions-for-xml-data

bgruening (commented):
@TanguyGen thanks a lot.

You might want to look at https://github.com/galaxyproject/galaxy-language-server its a nice extension for VisualStudioCode that helps you a lot with writing the wrappers.

tools/EMLassemblyline/eml2eal.xml
```diff
@@ -0,0 +1,53 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+       Rscript
+         '$__tool_directory__/eml2eal.R'
+         '$inputmetadatafile'
+    ]]></command>
+    <inputs>
+        <param name="inputmetadatafile" type="data" format="xml" label="EML metadata file" help="Insert only metadata using EML standard"/>
+    </inputs>
+    <outputs>
+        <collection name="EML_templates" type="list" label="EML Templates">
```
bgruening, line 15: How many outputs are generated? If only one XML file is ever generated, this is overkill and not needed. Simply add a `<data >` tag.

tools/EMLassemblyline/make_eml.xml
```diff
@@ -0,0 +1,70 @@
...
+        ln -s '$input' '${input.element_identifier}.txt' &&
+    #end for
+    find . -type f -empty -print -delete &&
+    Rscript '$__tool_directory__/make_eml.R'
+    ]]></command>
+    <inputs>
+        <param name="templates" type="data_collection"  collection_type="list" format="tabular" label="Upload all EAL templates" help= "Input a data collection." />
+    </inputs>
+    <outputs>
+        <data name="metadataout" format="xml" >
```
bgruening, line 20: How many output files are ever generated? If only one file, than this collection is overkill. Simply add a `<data>` tag without discover_datasets.

TanguyGen: Thanks a lot for the review !
I didn't know this extension existed but it seems super usefull thanks !
There is only one output but we don't know its name and by putting it into <data>instead of <collection> tag it does not create a collection does it ?
But there have been some other issues that I'm trying to correct, the tool now create  an empty XML that didn't appear before

bgruening: You can do the following at the end of your command section if you do not know the name.

```
mv output_template/*.pdf my_output.pdf
```

Then you know the output that you want to collect.

tools/EMLassemblyline/make_eml.xml
```diff
@@ -32,11 +32,13 @@
                         <element name="personnel" value="personnel.txt" />
                 </collection>
             </param>
-            <output name="metadataout" ftype="xml" count="1" >
-                <assert_contents>
-			<is_valid_xml />
-			<xml_element path="./dataset"/>
-                </assert_contents>
+            <output name="metadataout" count="1" >
+		<discovered_dataset designation="metadaout1" ftype="xml">
```
bgruening, line 36: this is invalid

try for example:

```
      <output name="metadataout" ftype="xml">
        <assert_contents>
          <is_valid_xml/>
        </assert_contents>
```

tools/EMLassemblyline/make_eml.xml
yvanlebras: Thank you Björn for this tip! Here, it is better to keep the original filename... I will try to see the best manner to do that

bgruening (approved):

bgruening:
Thanks.

yvanlebras:
Big big thank to @TanguyGen and you @bgruening !!!!! New tools for data/metadata management on Galaxy! And more to come !!!! This really give more interest in the zenodo - Galaxy recent dev! Really need to look at taking it as example to create DataONE oriented tools and others!

PR #82, opened 2023-11-17
update make_eml

PaulineSGN:


yvanlebras (approved):
THANK YOU so much Pauline !!!!!!!!

bgruening:
Now its too late, but you need to bump the galaxy0 version to galaxy1 in those cases. This is for reproducibility and tools versions.

yvanlebras:
oh, sorry....  We will pay more attention next time, ITM we was seeing a lack of functionnalities on the existing version so it seems best that this version replace the old one...

PR #86, opened 2023-12-22
Pauline sgn patch 1

PaulineSGN:


tools/EMLassemblyline/geo_cov_template.xml
```diff
@@ -1,6 +1,6 @@
 <tool id="geo_cov_template" name="Make geographic coverage template" version="0.1.0+galaxy0" profile="22.05">
     <description>for EML metadata creation</description>
-    <requirements>
+     <requirements>
```
bgruening, line 3: this is not using the container

tools/EMLassemblyline/make_eml.xml
```diff
@@ -1,134 +1,139 @@
...
+    </inputs>
+    <outputs>
+		 <data name="metadataout" from_work_dir="eml.xml" format="xml" label="EML"/>
+		 <data name="error" from_work_dir="err.txt" format="txt" label="Warning message"/>
+    </outputs>
+    <tests>
+        <test expect_num_outputs="2"> 
+            <param name="templates" value="annotations.txt,keywords.txt,attributes_data_blary_al.txt,catvars_data_blary_al.txt,geographic_coverage.tabular,personnel.txt" >
+            </param>
+            <output name="metadataout" >
```
bgruening, line 107: can you please check if the indentation is 4 spaces? something is off here

tools/EMLassemblyline/taxo_cov_template.xml
```diff
@@ -1,8 +1,7 @@
 <tool id="taxo_cov_template" name="Make taxonomic coverage template" version="0.1.0+galaxy0" profile="22.05">
```
bgruening, line 1: please make sure you increase all the tool versions - the suffix here to from 0 to 1

tools/EMLassemblyline/eal_table_template.xml
```diff
@@ -1,8 +1,7 @@
 <tool id="eal_table_template" name="Make templates for data tables" version="0.1.0+galaxy0" profile="22.05">
```
bgruening, line 1: increase tool version please

tools/EMLassemblyline/eal_table_template.xml
```diff
@@ -22,8 +21,8 @@
          <param name="inputdata" type="data_collection"  collection_type="list"  label="Upload all data files wich are data table" help= "Input a data collection."/>
     </inputs>
     <outputs>
-       <collection name="templates" type="list" label="Templates for data table">
-             <discover_datasets pattern="__designation_and_ext__"/>
+       <collection name="templates" type="list" label="Data table templates">
+             <discover_datasets pattern="(?P&lt;designation&gt;.+\.txt)" format="txt"/>
```
bgruening, line 25: Do you want to add a test here for this change?

bgruening (commented):
Thanks @PaulineSGN!

Can you remind me why the container is needed?

yvanlebras:
Hi Björn. Thank your for reviewing! We are using  container following our dedicated discussion because we are using a dev version of eml assembly line R package and it was not OK to use conda here. We will update through conda when the R package will be officially updated

tools/EMLassemblyline/geo_cov_template.xml
yvanlebras: Yes, because there is no change here and it appears there is an issue with this function on the container. Maybe the best is to remove this tool from the PR?

yvanlebras: But there is better descriptions so it seems OK to keep this tool on the PR but maybe not improve version

tools/EMLassemblyline/eal_table_template.xml
yvanlebras: Done. Thank you!

tools/EMLassemblyline/taxo_cov_template.xml
yvanlebras: Done. Thank you!

tools/EMLassemblyline/make_eml.xml
yvanlebras: I think it is ok now! Thank you Björn!!

tools/EMLassemblyline/eal_table_template.xml
yvanlebras: If ok for you we will see for further update of the tool and merge today this so I can try to test it on eu next week and share it to engineer we have who need to use it if possible

yvanlebras (approved):

PR #88, opened 2024-01-05
Run idw tool

MarkusKonk:


bgruening:
@MarkusKonk this will pass now. If you are fine with my changes we can merge this one.

The next step would be to number 3 here: https://galaxyproject.eu/posts/2020/08/22/three-steps-to-galaxify-your-tool/

PR #90, opened 2024-01-17
Add retrieving occurences data from obis

Marie59:
New tool to retrieve occurences data from Obis which can be chained with obisindicators tool. It relies on the package robis

### DIFF
--- a/tools/obisindicators/robis.xml
+++ b/tools/obisindicators/robis.xml
@@ -1,0 +1,94 @@
+<tool id="obis_data" name="OBIS occurences" version="@VERSION@" profile="20.01">
+    <description>retrieve data</description>
+    <macros>
+        <import>macro_obis.xml</import>
+    </macros>
+    <expand macro="topic"/>
+    <expand macro="requirements">
+        <requirement type="package" version="2.11.3">r-robis</requirement>
+    </expand>
+    <required_files>
+        <include type="literal" path="robis.r"/>
+    </required_files>
+    <command detect_errors="exit_code"><![CDATA[
+        Rscript
+            '$__tool_directory__/robis.r'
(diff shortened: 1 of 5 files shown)


tools/obisindicators/robis.xml
```diff
@@ -0,0 +1,94 @@
...
+            '$lat_min'
+            '$lat_max'
+            '$long_min'
+            '$long_max'
+            '$output'
+        ]]>
+    </command>
+    <inputs>
+        <param name="species" type="text" label="Scientific name of the species" help="Genus species format, eg : Scomber scombrus"/>
+        <param name="taxon" type="text" label="Taxon ID"/>
```
bgruening, line 27: should we restrict this text filed a bit? Ony letters? only letter and numbers?

Marie59: Yes that's a good idea !

Marie59: Should I use a sanitizer like the following for species :
`            <sanitizer invalid_char="">
                <valid initial="string.ascii_letters">
                </valid>
            </sanitizer>`

bgruening: Mh I guess a validator is needed

Marie59: Ah thanks !!

Marie59:
uh weird it does not like the validators ...

tools/obisindicators/robis.xml
bgruening: https://docs.galaxyproject.org/en/latest/dev/schema.html#id54

Marie59:
Youhou !! Thanks @bgruening  !!!

PR #96, opened 2024-02-24
Fix: offset issue in column access for medenv

jeremyfix:
Hello, sorry for this remaining issue which I probably introduced when changing the way to access the Latitude,Longitude, Depth, Date   fields; 

The test was successfully running because of a side effect. I was not giving to the test the column numbers as given by galaxy. 

@yvanlebras  : This is fixing the issue you noticed me about.


yvanlebras (approved):
THANK YOU SO MUCH Jérémy!!!!

bgruening:
Thanks 👍

PR #98, opened 2024-03-12
add otb_band_math tool

MarkusKonk:
This tool acts as a wrapper for an OGC API Process called OTB.BandMath. It takes the parameters from a user and triggers the service running outside Galaxy.


tools/ogcProcess_otb_bandmath/OTB_BandMath.xml
```diff
@@ -0,0 +1,91 @@
...
+        </citation>
+        <citation type="bibtex">
+            @Article{jsonlite,
+                title = {The jsonlite Package: A Practical and Consistent Mapping Between JSON Data and R Objects},
+                author = {Jeroen Ooms},
+                journal = {arXiv:1403.2805 [stat.CO]},
+                year = {2014},
+                url = {https://arxiv.org/abs/1403.2805},
+            }
+        </citation>
```
bgruening, line 89:
```suggestion
        <citation type="doi">10.48550/arXiv.1403.2805</citation>
```

tools/ogcProcess_otb_bandmath/OTB_BandMath.xml
```diff
@@ -0,0 +1,91 @@
...
+            <option value="float">float</option>
+            <option value="uint8">uint8</option>
+            <option value="uint16">uint16</option>
+            <option value="int16">int16</option>
+            <option value="int32">int32</option>
+            <option value="double">double</option>
+        </param>
+        <param type="select" name="output_type" label="Output image format" help="Select .png for obtaining the response as an OGC web service. Select .tiff or .jpeg for using it in a workflow.">
+            <option value="png">.png (for OGC web services)</option>
+            <option value="tiff">.tiff (for workflows)</option>
```
bgruening, line 35: I don't understand why tiff is good for workflow but png not?

tools/ogcProcess_otb_bandmath/OTB_BandMath.xml
```diff
@@ -0,0 +1,91 @@
...
+                --mathExpression '$math_expression'
+                --outputImage '$output_image' 
+                --outputType '$output_type' 
+                --outputFormat '$output_format' 
+                --outputData '$output_data'    
+        ]]>
+    </command>
+    <inputs>
+        <param type="data" format="txt" name="file" label="List of images" help="Image-list of operands to the mathematical expression (.txt file, one per line)" />
+        <param type="integer" name="processing_memory" label="Random-Access Memory (RAM)" value="256" optional="true" help="Available memory for processing (in MB)" />
```
bgruening, line 23: min/max should be added maybe?

tools/ogcProcess_otb_bandmath/OTB_BandMath.xml
```diff
@@ -0,0 +1,91 @@
...
+        <data name="output_data" format="png">
+            <change_format>
+                <when input="output_format" value="getUrl" format="txt" />
+                <when input="output_type" value="tiff" format="tiff" />
+                <when input="output_type" value="jpeg" format="jpeg" />
+            </change_format>
+        </data>
+    </outputs>
+    <tests>
+        <test>
```
bgruening, line 53: one test should be added for the getting a URL back I think.

tools/ogcProcess_otb_bandmath/OTB_BandMath.xml
bgruening: Ah do you mean if you want to pass it along to the next tool it strictly needs to be in tiff and jpeg, so the URL needs to be like that?

If so maybe we should make the output_format a conditional and offer different output_type options?

MarkusKonk: Right, if the tool should be used as input for another process, it has to be tiff or jpeg. Do you mean conditional by asking the user if the tool should be used in a workflow?

bgruening: Something like that:

```
<conditional name="outputs">
    <param type="select" name="output_format" label="Download the result to your Galaxy history or get the URL?" help="Select 'Get URL' to use the result as input for another OGC API Process in a workflow.">
        <option value="download">Download</option>
        <option value="getUrl">Get URL</option>
    </param>
    <when value="download">
        <param type="select" name="output_type" label="Output image format" help="Select .png for obtaining the response as an OGC web service. Select .tiff or .jpeg for using it in a workflow.">
            <option value="png">.png (for OGC web services)</option>
            <option value="tiff">.tiff (for workflows)</option>
            <option value="jpeg">.jpeg (for workflows)</option>
        </param>    </when>
    <when value="getURL">
        <param type="select" name="output_type" label="Output image format" help="Select .png for obtaining the response as an OGC web service. Select .tiff or .jpeg for using it in a workflow.">
            <option value="tiff">.tiff (for workflows)</option>
            <option value="jpeg">.jpeg (for workflows)</option>
        </param>
    </when>
</conditional>        
```

bgruening: If we do tools manually we can do those things and prevent user errors, I guess we can not do that when we convert automatically.

MarkusKonk: Mh, but what if a user wants to get the URL to the png file? With this conditional, this would not be possible. GetUrl is needed if the user wants to create a workflow but there might be cases other than creating workflows where the user wants the URL.

PR #99, opened 2024-03-12
Ogc process otb mean shift smoothing

MarkusKonk:
This tool acts as a wrapper for an OGC API Process called OTB.MeanShiftSmoothing. It takes the parameters from a user and triggers the service running outside Galaxy.

Note: It works in a similar way as the previous PR created for the tool OTB.BandMath. Here, max values are not possible for every numeric input.


tools/ogcProcess_otb_meanShiftSmoothing/OTB_MeanShiftSmoothing.xml
```diff
@@ -0,0 +1,116 @@
...
+    <description> smooths an image using the MeanShift algorithm</description>
+    <requirements>
+        <requirement type="package" version="4.3.1">r-base</requirement>
+        <requirement type="package" version="1.20.4">r-getopt</requirement>
+        <requirement type="package" version="0.2.3">r-httr2</requirement>
+        <requirement type="package" version="1.8.7">r-jsonlite</requirement>
+    </requirements>
+    <command detect_errors='exit_code'>
+        <![CDATA[
+            Rscript $__tool_directory__/OTB_MeanShiftSmoothing.R
```
bgruening, line 11:
```suggestion
            Rscript '$__tool_directory__/OTB_MeanShiftSmoothing.R'
```

bgruening (commented):
Test are unfortunately failing.

If you put all those tool into one directory you could share a lot of code via macros.

bgruening:
You can now see the test failures at https://github.com/galaxyecology/tools-ecology/actions/runs/8253335338?pr=99 scroll down please.

bgruening:
Also there is .. WARNING: Repository name [OTB.MeanShiftSmoothing] invalid. Repository names must contain only lower-case letters, numbers and underscore.

MarkusKonk:
Mhh, not sure what is meant by the warning. Which repository name should I change?

MarkusKonk:
no idea why lintr still complains. I don't get any issues locally and here I also don't get a concrete error message.

MarkusKonk:
is there a way to get a more concrete error message from the lintr? I don't why it still complains. I don't have any issues locally

bgruening:
I have only https://github.com/galaxyecology/tools-ecology/actions/runs/8264012065/job/22606726928?pr=99

We could disable the linting again if that is better.

MarkusKonk:
I did the linitng locally in RStudio using the lintr package. I don't see any issues anymore. So, if disabling it for this tool is an option, it would be helpful :)

bgruening:
You are the R expert, this is what we are using: https://github.com/galaxyecology/tools-ecology/blob/master/.github/styler.R

https://github.com/galaxyecology/tools-ecology/blob/master/.github/workflows/pr.yaml#L214

MarkusKonk:
I also used the R styler but it's still failing...I think disabling it is fine ;).

MarkusKonk:
It's a bit difficult to address the lintr issue if it doesn's provide more concrete descriptions of the issue.

bgruening (approved):

PR #101, opened 2024-03-20
Add moorev marine species recognition opencv pytorch

yvanlebras:


tools/Moorev/viame.xml
bgruening, line 1: that one should not be here I guess?

yvanlebras: this is a GxIT so I push it there to track addition of the toolsuite (classical Galaxy tools + Viame GxIT one) but I guess finally, this GxIT will be pushed somewhere else (directly on eu / org / fr usegalaxy repo). BUT we can keep it here to work on it no ? Duno what is the best

PR #105, opened 2024-03-22
Argo decoder tool

Marie59:
So, this is a batch tool using a docker image as a requirement. 
To work properly we will need to add a line there : https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/files/galaxy/tpv/interactive_tools.yml#L52 to change the user. 

The docker image is heavy so we'll see if it's possible to use it. 
The image once build is publicly accessible there  ghcr.io/euroargodev/coriolis-data-processing-chain-for-argo-floats:061a-galaxy 
For now the docker file is on a private gitlab but will soon be on a public github thanks to @leobr29.

I am starting this PR to see what needs to be improve and see the best ways to make this new tool work. This is a 1st version to make a demo for the Argo community in April. The aim is then to improve the tool (for instance remove all the zip and make more readable inputs and outputs) and create a workflow where we can access the argo data, decode them, and then visualize and analyze them with scoop-argo, ODV and DIVAnd.

I am not sure I'm making the PR in the right place. I put it with the other argo related tool.
Thoughts ?? Comments ??

Plus, I don't know how to make planemo test work (or if it's even possible) with a docker as a dependence


tools/ocean/argo_decoder.xml
```diff
@@ -0,0 +1,63 @@
...
+
+    ]]>
+    </command>
+     <environment_variables>
+        <environment_variable name="APP_HOME">/app</environment_variable>
+        <environment_variable name="APP_RUN_FILE">run_decode_argo_2_nc_rt.sh</environment_variable>
+    </environment_variables>
+    <inputs>
+        <param name="raw_data" type="data" format="zip" label="Archived input argo raw data"/>
+        <param name="float_wmo" type="text" label="Input WMO number" help="Input float WMO numer (example : 6903014)"/>
```
bgruening, line 36: This could be a float, isn't it?

tools/ocean/argo_decoder.xml
```diff
@@ -0,0 +1,63 @@
...
+        ##$APP_HOME/$APP_RUN_FILE /mnt/runtime 'rsynclog' 'all' 'configfile' '/app/config/_argo_decoder_conf_ir_sbd.json' 'xmlreport' 'argo-decoder-report.xml' 'floatwmo' '$float_wmo' 'PROCESS_REMAINING_BUFFERS' '1' &&
+        /app/run_decode_argo_2_nc_rt.sh /mnt/runtime 'rsynclog' 'all' 'configfile' '/app/config/_argo_decoder_conf_ir_sbd.json' 'xmlreport' 'argo-decoder-report.xml' 'floatwmo' '$float_wmo' 'PROCESS_REMAINING_BUFFERS' '1' &&
+
+        ## copy (because we need to be in working directory) and create output archive
+        cp -R /mnt/data/output/* . &&
+        zip -r argo-decoder-outputs.zip .
+
+    ]]>
+    </command>
+     <environment_variables>
```
bgruening, line 30: This is not needed anymore I assume? Its not used in the command line.

tools/ocean/argo_decoder.xml
```diff
@@ -0,0 +1,63 @@
+<tool id="coriolis_data_processing_chain_for_argo_floats" name="coriolis data processing" version="@VERSION@">
```
bgruening, line 1:
```suggestion
<tool id="coriolis_data_processing_chain_for_argo_floats" name="coriolis data processing" version="@VERSION@" profile="22.05">
```

tools/ocean/argo_decoder.xml
```diff
@@ -0,0 +1,63 @@
...
+        ## load inputs into data directory
+        cp '$raw_data' '/tmp/raw_data.zip' &&
+        unzip /tmp/raw_data.zip -d /mnt/data &&
+
+        ## execute decoder script
+        ##$APP_HOME/$APP_RUN_FILE /mnt/runtime 'rsynclog' 'all' 'configfile' '/app/config/_argo_decoder_conf_ir_sbd.json' 'xmlreport' 'argo-decoder-report.xml' 'floatwmo' '$float_wmo' 'PROCESS_REMAINING_BUFFERS' '1' &&
+        /app/run_decode_argo_2_nc_rt.sh /mnt/runtime 'rsynclog' 'all' 'configfile' '/app/config/_argo_decoder_conf_ir_sbd.json' 'xmlreport' 'argo-decoder-report.xml' 'floatwmo' '$float_wmo' 'PROCESS_REMAINING_BUFFERS' '1' &&
+
+        ## copy (because we need to be in working directory) and create output archive
+        cp -R /mnt/data/output/* . &&
```
bgruening, line 25: You could zip it in the /mnt dir and just copy the zip file to $outputs_data

tools/ocean/argo_decoder.xml
```diff
@@ -0,0 +1,63 @@
...
+        <token name="@VERSION@">061a</token>
+        <token name="@VERSION_DOCKER@">@VERSION@-galaxy</token>
+    </macros>
+    <requirements>
+        <container type="docker">ghcr.io/euroargodev/coriolis-data-processing-chain-for-argo-floats:@VERSION_DOCKER@</container>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        
+        ## prepare outputs directory
+        mkdir -p /mnt/data/output/log &&
```
bgruening, line 13: why do we need /mnt at all? Is this something the APP needs to?

tools/ocean/argo_decoder.xml
Marie59: yes we are still working on understanding how the environment var works here because when we defined them here it does not seem to know them.

tools/ocean/argo_decoder.xml
Marie59: I think the decoder need this tree structure for now. But that's a question more for @leobr29

tools/ocean/argo_decoder.xml
Marie59: even better it can be integer ;)

leobr29:
test

tools/ocean/argo_decoder.xml
bgruening: Maybe you ENV are not available for all users and the user which Galaxy is starting the container has no access to them?

tools/ocean/argo_decoder.xml
bgruening: `cp argo-decoder-outputs.zip '$outputs_data'` will work.

Or you can use `pushd` and `popd`

tools/ocean/argo_decoder.xml
bgruening: ok!

tools/ocean/argo_decoder.xml
leobr29: The environments variables are declared directly in the Dockerfile :

```Dockerfile
ENV APP_DIR=/app
ENV APP_RUN_FILE=run_decode_argo_2_nc_rt.sh
```

I tried to run the container with a random user on my computer :

```bash
docker run -it -u 123456:123456 ghcr.io/euroargodev/coriolis-data-processing-chain-for-argo-floats:061a-galaxy sh
$ id   
uid=123456 gid=123456 groups=123456
$ echo $APP_HOME/$APP_RUN_FILE             
/app/run_decode_argo_2_nc_rt.sh
$ printenv
HOSTNAME=307c9416801e
HOME=/
DATA_HOME=/mnt/data
MCR_CACHE_ROOT=/tmp/matlab/cache
REF_HOME=/mnt/ref
TERM=xterm
APP_HOME=/app
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
APP_DIR=/app
RUNTIME_HOME=/mnt/runtime
PWD=/app
APP_RUN_FILE=run_decode_argo_2_nc_rt.sh
```

Is it possible that it's a synthaxe problem in our configuration file ? I have the impression that when we use the `$` character there's an error even before the script runs.

tools/ocean/argo_decoder.xml
leobr29: ok we'll use '$outputs_data', thanks.

leobr29: @Marie59 can you replace next code ?
```bash
        cp -R /mnt/data/output/* . &&
        zip -r argo-decoder-outputs.zip .
```
by :
```bash
        cd /mnt/data/output &&
        zip -r argo-decoder-outputs.zip . &&
        cp argo-decoder-outputs.zip '$outputs_data'
``

File: tools/ocean/argo_decoder.xml (hunk 2 above)
Comment (Marie59) [reply tools/ocean/argo_decoder.xml:31]: @bgruening would you know what's the issue here ?

Comment (bgruening) [reply tools/ocean/argo_decoder.xml:31]: If you use a $env then you need to shield the env. Because in the template language the $ is a reserved keyword. Use \$env.

**Comment by Marie59** [comment]
@bgruening Is the tool good enough for a 1st try to deploy it or not ?

File: tools/ocean/argo_decoder.xml [hunk 6]
```diff
@@ -0,0 +1,76 @@
...
+        <param name="float_wmo" type="integer" label="Input WMO number" help="Input float WMO numer (example : 6903014)"/>
+    </inputs>
+    <outputs>
+        <data name="outputs_data" format="zip" from_work_dir="argo-decoder-outputs.zip" label="argo decoder netcdf outputs"/>
+    </outputs>
+    <tests>
+        <test expect_num_outputs="1">
+            <param name="raw_data" value="data_input_20240319.zip"/>
+            <param name="float_wmo" value="6903014"/>
+            <output name="outputs_data" value="argo-decoder-outputs.zip"/>
```
bgruening, line 41: what do you think about using https://docs.galaxyproject.org/en/latest/dev/schema.html#assertions-for-archives-zip-tar

To assert certain files in this archive, this would also enable us to remove this file.

tools/ocean/argo_decoder.xml
```diff
@@ -0,0 +1,76 @@
...
+        cp argo-decoder-outputs.zip '$outputs_data'
+
+    ]]>
+    </command>
+    <inputs>
+        <param name="raw_data" type="data" format="zip" label="Archived input argo raw data"/>
+        <param name="float_wmo" type="integer" label="Input WMO number" help="Input float WMO numer (example : 6903014)"/>
+    </inputs>
+    <outputs>
+        <data name="outputs_data" format="zip" from_work_dir="argo-decoder-outputs.zip" label="argo decoder netcdf outputs"/>
```
bgruening, line 35:
```suggestion
        <data name="outputs_data" format="zip" from_work_dir="argo-decoder-outputs.zip" label="Argo decoder netcdf outputs"/>
```

tools/ocean/argo_decoder.xml
```diff
@@ -0,0 +1,76 @@
...
+    ]]>
+    </help>
+    <citations>
+        <citation type="bibtex">
+            @Manual{,
+            title = {Coriolis data processing chain for Argo floats},
+            author = {Ifremer},
+            year = {2024},
+            note = {https://github.com/euroargodev/Coriolis-data-processing-chain-for-Argo-floats}
+        </citation>
```
bgruening, line 73:
```suggestion
```

The DOI is maybe enough?

tools/ocean/test-data/data_input_20240319.zip
bgruening, line 1: can we make this smaller?

bgruening:
.shed.yml is missing I think

PR #106, opened 2024-04-05
add SdmPredictors_list_layers tool

Rassine:
This tool helps you get one or multiple list of environemental layers from multiple sources, including :
WorldClim <https://www.worldclim.org/>
ENVIREM <https://envirem.github.io/>
Bio-ORACLE <https://bio-oracle.org/>
and MARSPEC <http://www.marspec.org/>


tools/sdmpredictors/sdmpredictors_list_layers.xml
```diff
@@ -0,0 +1,200 @@
+<tool id="sdmpredictors_list_layers" name="SdmPredictors List Layers" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
+    <description>Terrestrial and marine predictors for species distribution modelling, list climate layers from multiple datasets</description>
```
bgruening, line 2: I recommend to keep the description short and crisp. It's displayed in the tool bar and has not so much space

tools/sdmpredictors/sdmpredictors_list_layers.xml
```diff
@@ -0,0 +1,200 @@
...
+                        <param name="marine" value="false"/>
+                        <param name="freshwater" value="false"/>
+                    </section>
+                    <section name="advanced_options">
+                        <param name="monthly" value="true"/>
+                        <param name="version" value="0"/>
+                    </section>
+                </conditional>
+            </repeat>
+            <output_collection name="datasets_output" type="list" count="3"/>
```
bgruening, line 160: You could add here asserts to test the outputs.

bgruening: You might also want to test for the correct datatyp here. E.g. with ftype=

bgruening (commented):
Pretty cool and damn good. I added a few questions inline.

tools/sdmpredictors/sdmpredictors_list_layers.xml
yvanlebras: Done

tools/sdmpredictors/sdmpredictors_list_layers.xml
Rassine: Many thanks for your review! I added the changes you recommended. I do not exactly know witch tests would be pertinent to implement for a galaxy tool, by all mean don't hesitate to mention if you think somenthing is missing.

tools/sdmpredictors/sdmpredictors_list_layers.xml
```diff
@@ -0,0 +1,204 @@
...
+            ${s.choose_layer.environement.terrestrial}
+            ${s.choose_layer.environement.marine}
+            ${s.choose_layer.environement.freshwater}
+            ${s.choose_layer.advanced_options.monthly}
+            ${s.choose_layer.advanced_options.version}
+            #if str($s.choose_layer.input_layer) == 'layers_future'
+                ${s.choose_layer.advanced_options.scenario}
+                ${s.choose_layer.advanced_options.year}
+            #end if
+            #if str($s.choose_layer.input_layer) == 'layers_paleo'
```
bgruening, line 28: this could be an else if

tools/sdmpredictors/sdmpredictors_list_layers.xml
```diff
@@ -0,0 +1,204 @@
...
+                        <param name="freshwater" value="true"/>
+                    </section>
+                    <section name="advanced_options">
+                        <param name="monthly" value="true"/>
+                        <param name="version" value="0"/>
+                    </section>
+                </conditional>
+            </repeat>
+            <output_collection name="datasets_output" type="list" count="3">
+                <element name="1_data_paleo" file="1_data_paleo.tsv" ftype="tsv"/>
```
bgruening, line 161: you can now use something like this here: https://github.com/galaxyproject/tools-iuc/blob/01e7e0814f8bb4253a92ee4541b555e9af74fb0a/tools/checkm/tree.xml#L88

See also here. https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents

bgruening (commented):
Two more comments inline :)

One thing that I do not know but I wanted to point out is, that TSV != tabular in Galaxy. `tabular` is preferred imho.

yvanlebras:
THANK YOU Björn! Considering the tabular/tsv issue, it seems to me tabular is only used on Galaxy side, so for me, it would be better to use a datatype known externally from Galaxy, as tsv I guess. Moreover, it seems to me tsv and tabular are exactely the same content and so, Galaxy can maybe consider tabular and tsv datatype the same way, and particularly for Galaxy tools taking tabular as input. Do you think this makes sense and maybe we have to initiate a work on the main Galaxy code repo ?

bgruening:
> THANK YOU Björn! Considering the tabular/tsv issue, it seems to me tabular is only used on Galaxy side, so for me, it would be better to use a datatype known externally from Galaxy, as tsv I guess. Moreover, it seems to me tsv and tabular are exactely the same content and so, Galaxy can maybe consider tabular and tsv datatype the same way, and particularly for Galaxy tools taking tabular as input. Do you think this makes sense and maybe we have to initiate a work on the main Galaxy code repo ?

It has nothing to do with outside or inside Galaxy. Its unfortunately just a historical issue that at some point someone has introduced tsv to Galaxy and made everything very confusing. `tabular` is just the name of a tabular formated file (with or without header), you can call it tsv if you like but we need to refer to as `tabular` in the Galaxy naming schema. This does not mean you need to add extensions to your files. Extensions make it easier but you can also have `foo.tsv` files recognised as `tabular` if you like. I don't know if we can fix this mess. pinging @bernt-matthias if he has something to correct/add.

bgruening (approved):
This is pretty cool @Rassine! Sorry for keeping you up at night with the tsvs.

bernt-matthias:
Not much to add. The data is the same: tab separated data in columns. 

To my understanding `tabular` is more flexible towards problems in the file because `tsv` makes more assumptions (e.g. it assumes a header). IMO having header metadata automatically is a nice feature that I miss for tabular.  I was wondering  often if we should use tsv more often if we know that the data behaves well .. but I never made up my mind. But it will be a long and bumpy way because users will need change datatype tabular <-> tsv manually very often.

Rassine:
> This is pretty cool @Rassine! Sorry for keeping you up at night with the tsvs.

haha don't worry about it! But I was quite surprise to see that while running  `planemo t ` my test failed because the only file type recognised with `ftype="tabular" ` was `.tabular` extensions. Galaxy doesn't seem to recognise `.tsv` files as tabular, and if I specify a `.tsv` output file as tabular in the `<discover_datasets/>` flag , the `.tsv` extension is replaced by `.tabular` when downloading it from galaxy.

Whilst I don't mind using either tabular or tsv, I belive tsv is a more recognisable and easy to use format for the wider community.

PR #110, opened 2024-04-29
Ogc api processes wrapper

MarkusKonk:
This tool is wrapper for OGC API Processes - remote services that can be executed via a URL. The XML currently wraps around 700 processes from one server. More servers and more processes can be added using the tool https://github.com/AquaINFRA/OGC-API-Process2Galaxy. Adding/removing services requires another PR to Galaxy.
Planemo test was successful but the linter raised some warnings such as `WARNING: Param input [elev.dem] is not a valid Cheetah placeholder.`. The warning disappeared when I replaced the dot by an underscore. However, then the requests to the remote processes fail as they need the exact names as specified in the process description. Hence, I had to keep the input names as they are.


MarkusKonk:
@bgruening I wasn't aware that there is a file size limit. The xml is around 1,8MB. Do you have a suggestion?

bgruening:
We can relax the filesize limit :) we never had such a big XML file. This was meant for test data :)

bgruening:
> WARNING: Param input [elev.dem] is not a valid Cheetah placeholder.

This is a valid warning. You should not do `$elev.dem` instead you can do `${elev.dem}`. Does that work?

tools/aquainfra_importer/aquainfra_importer.xml
```diff
@@ -0,0 +1,13 @@
+<tool id="aquainfra_importer" name="AquaINFRA Importer" tool_type="data_source" version="1.0" profile="22.05">
```
bgruening, line 1: can you split this please into several PRs? Or is this just a bad branch with things that don't belong together?

MarkusKonk: oh, that is a mistake

MarkusKonk:
> > WARNING: Param input [elev.dem] is not a valid Cheetah placeholder.
> 
> This is a valid warning. You should not do `$elev.dem` instead you can do `${elev.dem}`. Does that work?

I will check it.

bgruening:
@MarkusKonk have a look at https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-inputs

Galaxy will write all input-vars in a json, which then your R code can read and execute. This should make the command section trivial and way less code to carry around. It will also make your R script a bit more testable imho.

bgruening:
I guess one tool per server is enough? So you can avoid the first conditional and nesting?

bgruening:
* Do we need things like "HelloPy" or "echo"?
* can we not put `optional="false"` into the XML is default and will safe a few bytes.

```
						<param name="out" label="out" optional="false" value="uint8" help="Output image" type="select">
							<option value="uint8">uint8</option>
							<option value="uint16">uint16</option>
							<option value="int16">int16</option>
							<option value="int32">int32</option>
							<option value="float">float</option>
							<option value="double">double</option>
						</param>
```

Is used 70 times, can we put this into a macro file? You can parameterize macros ([example](https://github.com/galaxyproject/tools-iuc/blob/main/tools/gemini/gemini_macros.xml#L99)  `@ARGUMENT@` belongs to the argument in token_argument).

bgruening:
out of interest is the tool form slow to load?

MarkusKonk:
> out of interest is the tool form slow to load?

Locally it was slowly, yes. I was hoping it's getting better on the server but if not, we will need to find a better solution.

MarkusKonk:
> I guess one tool per server is enough? So you can avoid the first conditional and nesting?

We will add more servers during the next few weeks.

bgruening:
I think its slow, because the webbrowser needs to render 1000 elements, so nothing the server can influence. At least that is my hunch. That also why I was asking to have per-server tools? To keep it a bit more reduced?

MarkusKonk:
> * Do we need things like "HelloPy" or "echo"?
> 
>     * can we not put `optional="false"` into the XML is default and will safe a few bytes.

I don't get how I can make this default?

bgruening:
`optional="false"` is the default. Just remove all strings `optional="default"`.

MarkusKonk:
> The exact same warning? it should at least have changes? Can you push your changes? And the warning?

I pushed an update but I am not yet done with all changes.

bgruening:
> If I remove the optional=false element from the process in the screenshot, then the input parameter becomes optional. So in this case it isn't the default.

That would be new, it should not be the case. We never write optional="false".

bgruening:
`If false, parameter must have a value. Defaults to false except when the type attribute value is select and multiple is true.`
https://docs.galaxyproject.org/en/latest/dev/schema.html#id41

MarkusKonk:
> `If false, parameter must have a value. Defaults to false except when the type attribute value is select and multiple is true.` https://docs.galaxyproject.org/en/latest/dev/schema.html#id41

Oh f&%$§§, my bad!

MarkusKonk:
I need to add optional=false for text input parameters. Otherwise it's optional. Or am I still doing something wrong?
image
image

bgruening:
If you really need a value in this field and a empty field is not valid, yes then please add it.

MarkusKonk:
Alright, here is the update. Changes can be seen in the commit plus I removed the server. However, the linter still complains about the parameters and outputs a warning for every input parameter having a "." in the name: `WARNING: Param input [type.linear.gamma] is not a valid Cheetah placeholder.`

bgruening:
I can see that the dot-syntax can be confusing with the nested, conditional/section syntax. Is it a big problem to name them `_` and not `.`?

All params with only true/false you could model as boolean-params, I think that would be a nice improvement.

MarkusKonk:
> I can see that the dot-syntax can be confusing with the nested, conditional/section syntax. Is it a big problem to name them `_` and not `.`?

The problem is that I take the names from the process description and need to name the parameters as defined there to build the requests.

bgruening:
Only the param names. So something like `--foo.bar.baz ${foo_bar_baz}`

MarkusKonk:
Ah I see, I will try.

MarkusKonk:
> All params with only true/false you could model as boolean-params, I think that would be a nice improvement.
Do you remember where you saw it? There should be boolean types with truevalue/falsevalue.

MarkusKonk:
Just the file size issue is remaining. Looks like it's working :).

bgruening:
> @MarkusKonk have a look at https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-inputs
> 
> Galaxy will write all input-vars in a json, which then your R code can read and execute. This should make the command section trivial and way less code to carry around. It will also make your R script a bit more testable imho.

Have you considered this one? It should make this tool and all following ones much much easier and will half the file size as well I assume.

MarkusKonk:
> > @MarkusKonk have a look at https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-inputs
> > Galaxy will write all input-vars in a json, which then your R code can read and execute. This should make the command section trivial and way less code to carry around. It will also make your R script a bit more testable imho.
> 
> Have you considered this one? It should make this tool and all following ones much much easier and will half the file size as well I assume.

I don't understand this one...

MarkusKonk:
Ah ok ok, I think I got it. Needs some rewriting but it should be done by tomorrow.

bgruening:
:) yeah, has nothing to do with generating XML. It contains all params in a structured, parsable way - JSON. Means your R script can just read a JSON file instead of creating a complicated CLI just to pass parameters over to R. That should make stuff much simpler in the XML (the command section becomes a one liner) but also your R script does not need to handle any command-line arguments anymore. Instead, you consume a JSON.

MarkusKonk:
> Only the param names. So something like `--foo.bar.baz ${foo_bar_baz}`

That worked by the way!

MarkusKonk:
> :) yeah, has nothing to do with generating XML. It contains all params in a structured, parsable way - JSON. Means your R script can just read a JSON file instead of creating a complicated CLI just to pass parameters over to R. That should make stuff much simpler in the XML (the command section becomes a one liner) but also your R script does not need to handle any command-line arguments anymore. Instead, you consume a JSON.

That was a great hint. This saved us around 5000 lines. Thanks!

bgruening:
You can probably safe the rest of the cheetah as well, isn't it? The algorithm/function should be in the JSON. The inputs as well? If inputs are not in there, I think there is a argument on the configfiles that can include inputs. And last but not least to output folder you simply hardcode to ./outputs/ and then use fro_dir when you collect the outputs.


Do you mind adding a few manual tests to the macro? Maybe some more complicated functions? I guess then we are ready.

MarkusKonk:
Do you mean `from_work_dir` for the outputs??

bgruening:
Not 100% sure and on vacation only with mobile ATM. The disxover_collection things can discover from a directory. But not sure what the name was. Maybe just directory? Please look at the schema page.

MarkusKonk:
> You can probably safe the rest of the cheetah as well, isn't it? The algorithm/function should be in the JSON. The inputs as well? If inputs are not in there, I think there is a argument on the configfiles that can include inputs. 

Oh wow, how could I oversee this. I don't even need a single if statement in the command... :D

MarkusKonk:
How about the following: I will anyway need to update the tool in two weeks after it's published on Galaxy based on the feedback I will get from the partners. We take the tool as it is now to have it available from tomorrow. In the meantime, I will work on the remaining open issues (output data, tests) and add it to the tool with the next PR. So you can enjoy your well-deserved holidays and I save a week :). Plus I am pretty curious regarding the performance on Galaxy.

bgruening (approved):

bgruening:
@electronicblueberry is looking at the performance. It's okish in Chrome and horrible in FF. There is hope that we can enhance this.

bgruening:
@MarkusKonk since this is just one server now. The tool, folder and ID should be renamed isn't it? Please note that the ID needs to be stable.

MarkusKonk:
Good point. I will change it now.

bgruening (approved):

bgruening:
Sorry, we have an other problem: `parameter 'channels_rgb_red': the attribute 'value' must be set for non optional parameters`. I think you need to add `value=""` for those params.

PR #113, opened 2024-05-08
Ogc api processes wrapper for the zoo project

MarkusKonk:


tools/zoo_project_ogc_api_processes/macros.xml
```diff
@@ -37,8 +37,12 @@
                 <param name="out" value="float"/>
                 <param name="ram" value="256"/>
                 <param name="exp" value="im1b3,im1b2,im1b1"/>
-                <param name="out_outformat" value="image/png"/>
-                <output name="output_data" file="otb_band_math_test_output.txt" compare="sim_size" delta="300" />
+                <param name="out_outformat" value="image/tiff"/>
+                <output name="output_data" file="otb_band_math_test_output.txt">
```
bgruening, line 41: If you specify a file here, galaxy will compare the file as diff plus do the asserts.

MarkusKonk: Awsome, thanks! Newest commit has a reduced number of processes (~100). The people in the project start preferring tools that are more categorised and do not contain the whole set of processes. This would probably reduce our performance issue in many cases. In this PR, for example, there are only OTB processes.

PR #116, opened 2024-06-28
Fix urls

Marie59:
Fixing some bad urls in shed.yml (mine did a bit of a mess now it's better) and @MarkusKonk for the `remote_repository_url` be careful I think your putting some other folder in the link you need to put the same than where is the .shed.yml file of your tool in the ecology github repo

### DIFF
--- a/tools/champ_blocs/.shed.yml
+++ b/tools/champ_blocs/.shed.yml
@@ -1,7 +1,8 @@
 categories:
     - Ecology
 owner: ecology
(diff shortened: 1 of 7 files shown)


bgruening:
Thanks! You are now triggering old test, which is good but also yields to new lint warnings :)

>.. WARNING (TestsExpectNumOutputs): Test 1: should specify 'expect_num_outputs' if outputs have filters

Marie59:
Oh nooooo !! I will update everything soon then ;)

PR #122, opened 2024-07-30
Terrabyte ogc api process

MarkusKonk:
This is a wrapper for an OGC API Process hosted on Terrabyte server. Don't worry about the cookie in the test. It's not valid but the process checks for a cookie that has the right format. So just passing an empty string or "abcd" is not possible.

### DIFF
--- a/tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
+++ b/tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
@@ -14,11 +14,11 @@
 		<inputs name="inputs" filename="inputs.json" data_style="paths"/>
 	</configfiles>
 	<inputs>
-		<param name="cookie" label="Cookie" optional="false" value="" type="text" help="See 'Help' for for more information regarding the cookie."/>
-		<param name="date" label="Date" optional="false" help="Center date for data query" type="text"/>
-		<param name="delta" label="Delta" optional="false" help="Days delta +- around the specified date" type="text"/>
-		<param name="aoi" label="AOI" optional="false" help="Area of interest as GeoJSON polygon to find intersection of products" type="text"/>
-		<param name="cloud_cover" label="Cloud Cover" optional="false" help="Maximum cloud coverage permitted for query" type="text"/>
+		<param name="cookie" label="Cookie" optional="false" value="" type="text"/>
+		<param name="date" label="Date" optional="false" help="Center date for data query in the format YYYY-MM-DD, e.g., '2019-06-29'." type="text"/>
+		<param name="delta" label="Delta" optional="false" help="Days delta +- around the specified date." type="text"/>
+		<param name="aoi" label="AOI" optional="false" help="Area of interest as Well-Known-Text (WKT) polygon to find intersection of products, e.g., 'POLYGON ((-7.664532 38.543869, -7.689248 38.141037, -7.159228 38.151837, -7.115289 38.554609, -7.664532 38.543869))'." type="text"/>
+		<param name="cloud_cover" label="Cloud Cover" optional="false" help="Maximum cloud coverage permitted for query, e.g. '0.1'." type="text"/>
 	</inputs>
 	<outputs>
 		<data name="output_data" format="txt" label="Water quality EO Application Package"/>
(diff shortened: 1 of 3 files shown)


tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
```diff
@@ -0,0 +1,49 @@
...
+			<param name="aoi" value="POLYGON ((-7.664532 38.543869, -7.689248 38.141037, -7.159228 38.151837, -7.115289 38.554609, -7.664532 38.543869))"/>
+			<param name="cloud_cover" value="0.1"/>
+			<assert_stderr>
+				<has_text text="HTTP 401 Unauthorized"/>
+			</assert_stderr>
+		</test>
+	</tests>
+	<help>
+		This process takes a Sentinel-2 Scene as input, and calculates chlorophyll-a concentration and turbidity in water bodies in an EO Application Package. 
+		It returns the results as tiffs representing the input scene. You need to have an account at the Leibniz Supercomputing Center (LRZ) with a grant for terrabyte.
```
bgruening, line 40:
```suggestion
		It returns the results as tiffs representing the input scene. You need to have an account at the Leibniz Supercomputing Center (LRZ) with a grant for "terrabyte". 
```

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
```diff
@@ -0,0 +1,49 @@
...
+			<param name="cloud_cover" value="0.1"/>
+			<assert_stderr>
+				<has_text text="HTTP 401 Unauthorized"/>
+			</assert_stderr>
+		</test>
+	</tests>
+	<help>
+		This process takes a Sentinel-2 Scene as input, and calculates chlorophyll-a concentration and turbidity in water bodies in an EO Application Package. 
+		It returns the results as tiffs representing the input scene. You need to have an account at the Leibniz Supercomputing Center (LRZ) with a grant for terrabyte. 
+		As part of the OGC Open Science Persistent Demonstrator you can ask for an account at LRZ/terrabyte.
```
bgruening, line 41:
```suggestion
```

Thos dynamic information we should add via https://galaxyproject.org/news/2024-06-05-help-forum-integration/
The demonstrator will disappear but the tool will stay.

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
```diff
@@ -0,0 +1,49 @@
...
+	</requirements>
+	<command detect_errors="exit_code"><![CDATA[
+	Rscript '$__tool_directory__/terrabyte_ogc_api_process_wrapper.R'
+		--outputData '$output_data'
+]]></command>
+	<configfiles>
+		<inputs name="inputs" filename="inputs.json" data_style="paths"/>
+	</configfiles>
+	<inputs>
+		<param name="cookie" label="Cookie" optional="false" value="" type="text" help="See 'Help' for for more information regarding the cookie."/>
```
bgruening, line 17: There is no help about this.

Is there no other way. This seems to be a general problem which we need to address.

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
```diff
@@ -0,0 +1,49 @@
...
+		<param name="date" label="Date" optional="false" help="Center date for data query" type="text"/>
```
bgruening, line 18: Wasn't there a data-picker already? Or was that on our todo list?

How should a date look like? Which format?

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
```diff
@@ -0,0 +1,49 @@
...
+		<param name="delta" label="Delta" optional="false" help="Days delta +- around the specified date" type="text"/>
```
bgruening, line 19: Why is this not an integer with min=1?

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
```diff
@@ -0,0 +1,49 @@
...
+		<param name="aoi" label="AOI" optional="false" help="Area of interest as GeoJSON polygon to find intersection of products" type="text"/>
```
bgruening, line 20: Does a normal user know how the input need to look like here? I would not :(

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
```diff
@@ -0,0 +1,49 @@
...
+		<param name="cloud_cover" label="Cloud Cover" optional="false" help="Maximum cloud coverage permitted for query" type="text"/>
```
bgruening, line 21: float?

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
MarkusKonk: Cookies/Tokens and in general authentication are a problem in our project. I hope this will be addressed in the second phase of the project...

bgruening: But realistically, how likely is it that someone can paste a cookie in here? And even worse this cookie is saved in the DB - if you share your history someone else has access to it.

MarkusKonk: The cookies expire after a few minutes. 
I am also not very happy with this cookie issue, but this is nothing we can solve until the end of the first project phase, which will be in six weeks. They are also aware of this issue and I hope that this demonstrator will provide some motivation to work in it during the next months.

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
MarkusKonk: I am not aware of a date picker but added some explanation for the format.

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
MarkusKonk: because the process description defines this parameter as string

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
MarkusKonk: added some explanation

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
MarkusKonk: same here, the process description defines this parameter as string

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
MarkusKonk: though it doesn't matter what the description says..., you're right, I changed it

tools/terrabyte_ogc_api_processes/terrabyte_ogc_api_process_wrapper.xml
MarkusKonk: you're right, changed it

bgruening:
Is that one ready?

MarkusKonk:
Ready!

PR #124, opened 2024-08-06
Change divan output label

Marie59:
I discovered a tiny glitch when trying to build my workflow this should fix it


tools/ocean/divandfull.xml
```diff
@@ -58,7 +58,7 @@
         <param name="selmax" type="integer" min="0" max="100" optional="true" value="0" label="Maximum of the salinity"/>
     </inputs>
     <outputs>
-        <data name="output_netcdf" label="DIVAnd netcdf output" from_work_dir="outputs_netcdf/*.nc" format="netcdf"/>
+        <data name="output_netcdf" label="DIVAnd_output.nc" from_work_dir="outputs_netcdf/*.nc" format="netcdf"/>
```
bgruening, line 61: You rely on the Name in your workflow?

tools/ocean/divandfull.xml
```diff
@@ -2,7 +2,7 @@
     <description>Data-Interpolating Variational Analysis in n dimensions</description>
     <macros>
         <token name="@TOOL_VERSION@">0.1.0</token>
-        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@VERSION_SUFFIX@">0.0.1</token>
```
bgruening, line 5: its a simple integer, 0 ... 1... 2...3

Marie59: Oh I thought it follows the rule when you have minor fixes 0.0.+1, update 0.+1 or major changes +1

tools/ocean/divandfull.xml
bgruening: But the ODV tool could just `ln -s $input ./foo.nc` and it would solve it no?

PR #130, opened 2024-09-17
Update brt tool restore r version for dismo package

PaulineSGN:


yvanlebras:
Thank you @PaulineSGN ! @bgruening it seems to me Pauline is trying to update the tool so it can be launched in eu, https://ecology.usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/ecology/ecoregion_brt_analysis/ecoregion_brt_analysis/0.1.0+galaxy0 but in fact, it seems to me the tool is ok regarding previous local and github tests, so maybe in fact, we don't need to retrograde the R version as proposed here and maybe more look at why it is not working on eu. You think you can look at it ?

Cheers !

bgruening:
I also don't think this PR is needed. What is the problem on EU?

yvanlebras:
we have this error :
```
Error in library(dismo, warn.conflicts = FALSE) :    there is no package called ‘dismo’ Execution halted
```
So it seems to me the container is not ok...

bgruening:
Please try again!

yvanlebras:
It is working properly!!! THANK YOU wonder Björn !

PR #132, opened 2024-10-09
first set of aquainfra processes

MarkusKonk:


tools/aquainfra_ogc_api_processes/aquainfra_ogc_api_processes.xml
```diff
@@ -0,0 +1,94 @@
...
+			</when>
+			<when value="peri-conv">
+				<param name="input_data" label="Table to be grouped by date, with date colum" help="URL to the table with a column containing a date. It can have other columns which will not be changed during this process." type="data" format="txt"/>
+				<param name="colname_date" label="Date column name" optional="false" help="Column name of the date column in the format defined above. Example: 'visit date'." type="text"/>
+				<param name="date_format" label="Date format" optional="true" help="The date format used to parse the date, i.e. to extract day, month and year from the date column, e.g.  or 'y-m-d' for dates like '1998-08-22' (this is the default) or 'y/m/d' for dates like '1998/08/22'." type="text"/>
+				<param name="group_to_periods" label="Periods to group the data into" optional="true" help="Define the periods that you want the data to be grouped into. Please follow the example: 'Dec-01:Mar-01,Mar-02:May-30,Jun-01:Aug-30,Sep-01:Nov-30' (first three letters of each month, then a minus/hyphen, then the day (two digits), then comma, then the next period)." type="text"/>
```
bgruening, line 36: what about a validator here? Maybe even a regex validator?

tools/aquainfra_ogc_api_processes/aquainfra_ogc_api_processes.xml
```diff
@@ -0,0 +1,94 @@
...
+				<param name="group_labels" label="Period labels" optional="true" help="Define names for the periods defined above, separated by a comma. Example: 'winter,spring,summer,autumn'." type="text"/>
+				<param name="year_starts_at_Dec1" label="Include December into next year" help="Tell whether the years in your grouping starts at first of December ('true') or not ('false'). This has to be reflected in the period definitions, these should not contradict each other." type="boolean" truevalue="true" falsevalue="false" checked="true"/>
```
bgruening, line 38:
```suggestion
				<param name="year_starts_at_dec1" label="Include December into next year" help="Tell whether the years in your grouping starts at first of December ('true') or not ('false'). This has to be reflected in the period definitions, these should not contradict each other." type="boolean" truevalue="true" falsevalue="false" checked="true"/>
```

tools/aquainfra_ogc_api_processes/aquainfra_ogc_api_processes.xml
```diff
@@ -0,0 +1,94 @@
...
+			</when>
+			<when value="mean-by-group">
+				<param name="input_data" label="Input table" optional="false" help="URL to the input table containing group identifier(s) and a value column. Groups are defined by one or more columns. Other columns present in the table will be removed in the process, retaining only those specified in cols_to_group_by and value_col. For example, use the result table from peri_conv. " type="data" format="txt"/>
+				<param name="colnames_to_group_by" label="Column names identifying group" optional="false" help="One or more column names identifying the group. A combination of all specified columns will be used to define unique groups for calculating the average value. Example: 'longitude, latitude, Year_adj_generated, group_labels, HELCOM_ID'. Another example: 'Year_adj_generated, group_labels, HELCOM_ID'." type="text"/>
```
bgruening, line 42: Just as an info, sorry if I told this already multiple times, if you would use Galaxy datasets here, you could use https://docs.galaxyproject.org/en/latest/dev/schema.html#data-column and everything will be super simple for your users.

tools/aquainfra_ogc_api_processes/aquainfra_ogc_api_processes.xml
```diff
@@ -0,0 +1,94 @@
...
+				<param name="colname_region_id" label="Column name of region identifier" optional="false" help="The name of the column containing identifiers (e.g., basin name) to distinguish the polygons on the map if multiple regions are present." type="text"/>
+			</when>
+		</conditional>
+	</inputs>
+	<outputs>
+		<data name="output_data" format="txt" label="$select_process"/>
+	</outputs>
+	<expand macro="tests"/>
+	<help>
+		Use the dropdown menu at the top to select the OGC API process and then complete the corresponding form to run the service.
```
bgruening, line 91: Maybe something about this specific OGC service? A URL etc...

bgruening (commented):
Do you want to add a creator tag and ack aquainfra? https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-creator

having 4-spaces indentation would make it so much more readable :)

tools/aquainfra_ogc_api_processes/aquainfra_ogc_api_processes.xml
```diff
@@ -0,0 +1,94 @@
...
+				<param name="input_data" label="Table to be merged with study region" optional="false" help="URL to the input table containing the in-situ data points with coordinates. Can be provided as Excel file or CSV file (comma-separated text file). The coordinates have to be in WGS84 coordinate system." type="data" format="txt"/>
+				<param name="colname_long" label="Column name for longitude" optional="false" help="Name of the column that contains longitude values (in WGS84)." type="text"/>
+				<param name="colname_lat" label="Column name for latitude" optional="false" help="Name of the column that contains latitude values (in WGS84)." type="text"/>
+			</when>
+			<when value="peri-conv">
+				<param name="input_data" label="Table to be grouped by date, with date colum" help="URL to the table with a column containing a date. It can have other columns which will not be changed during this process." type="data" format="txt"/>
```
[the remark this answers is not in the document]
MarkusKonk: It's not a URL passed as a string parameter, but a URL stored in a .txr file. It's the same process as in the OGC process. I will reformulate it to make that clearer.

tools/aquainfra_ogc_api_processes/aquainfra_ogc_api_processes.xml
MarkusKonk: This would be possible if we would pass the input file directly, but since we use .txt files to do that, I don't think we have that option.

tools/aquainfra_ogc_api_processes/aquainfra_ogc_api_processes.xml
MarkusKonk: tried but failed :/

MarkusKonk: At leaast, this one did not work: 
`Parameter group_to_periods: Value 'Dec-01:Mar-01,Mar-02:May-30,Jun-01:Aug-30,Sep-01:Nov-30' does not match regular expression ' ^(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)-[0-9][0-9]:(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)-[0-9][0-9](,(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)-[0-9][0-9]:(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)-[0-9][0-9])*$ '`

bgruening: Mh, seems to work https://regex101.com/r/TOsg0G/1

Maybe remove the trainling space?

MarkusKonk: That was the problem. I oversaw the spaces.

bgruening (approved):

PR #144, opened 2024-10-30
Copernicus marine minor fix AND ....

Marie59:
So, I noticed some really minor fixes to make while testing the tool

BUT I also noticed it's not working ... It seems that even thought I put my credentials in my preferences it's not getting them.

`
        #set $cmems_username = $__user__.extra_preferences.get('cmems_username', "")
        #set $cmems_password = $__user__.extra_preferences.get('cmems_password', "")
`

I guess there is something not working there. Not sure but could it be that this cmems_username and cmems_password should be in capital ? 
Do you have the PR of when this was implemented into the Galaxy preference user so I can have a look at it ?


Marie59:
Apparently there's still the same github security issue

bgruening:
@Marie59 you need to create for this `special` tool PRs from within the repo - not your fork.

PR #145, opened 2024-10-30
Copernics credentials issue

Marie59:
I also noticed it's not working ... It seems that even thought I put my credentials in my preferences it's not getting them.

`
#set $cmems_username = $__user__.extra_preferences.get('cmems_username', "") 
#set $cmems_password = $__user__.extra_preferences.get('cmems_password', "")
`

I guess there is something not working there. Not sure but could it be that this cmems_username and cmems_password should be in capital ?
Do you have the PR of when this was implemented into the Galaxy preference user so I can have a look at it ?

### DIFF
--- a/tools/ocean/copernicusmarine.xml
+++ b/tools/ocean/copernicusmarine.xml
@@ -9,8 +9,8 @@
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
     #import re
-        #set $cmems_username = $__user__.extra_preferences.get('cmems_username', "")
-        #set $cmems_password = $__user__.extra_preferences.get('cmems_password', "")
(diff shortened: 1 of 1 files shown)


Marie59:
so something weird is happening here. If I had to guess I think with github the os.getenv works however once on galaxy it does not and this that should work $__user__.extra_preferences.get('cmems_username', "") is not

yvanlebras:
Hi Marie! Thank you for digging into it! Duno if you message is for me and/or Anne and/or someone else ;) But definitely interested to help there!

Marie59:
Hi Yvan ! It's for anyone who can help/answer ;)

Marie59:
@bgruening Is there sommething I can do here ?

PR #150, opened 2024-12-20
Canyonb neural network for the ocean tool

Marie59:
Hi all !!

Christmas present a new tool for ocean analysis based on a neural network ! 

/!\ This is not to be merged until the to do is all checked by me ;)

TO DO

- [x] Add help for parameters
- [x] Add entire help section
- [x] Do a 2nd real test with real data (for now I tested  with some random input)
- [x] Check with Catherine for the min/max values possible for each parameters

Thank you for the up coming reviews !! 🎄​🎁​🎅​


tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,91 @@
+<tool id="bgc_canyon_b" name="Canyon-B" version="@VERSION@">
```
bgruening, line 1: Please add a peofile

tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,91 @@
...
+    <description>Robust Estimation of Open Ocean CO2 Variables and Nutrient Concentrations From T, S, and O2 Data Using Bayesian Neural Networks</description>
```
bgruening, line 2: Shorter description

tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,91 @@
...
+    <macros>
+        <token name="@VERSION@">0.9.3</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="4.3.2">r-base</requirement>
```
bgruening, line 7: r-base should not be needed, let's try without

tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,91 @@
...
+        <requirement type="package" version="@VERSION@">r-canyonb</requirement>
+    </requirements>
+    <required_files>
+        <include type="literal" path="canyonb.r"/>
```
bgruening, line 11: What is type literal?

tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,91 @@
...
+            '$variables'
+            '$epressure'
+            '$etemperature'
+            '$epsalinity'
+            '$edissoxygen'
+            '$output'
+        ]]>
+    </command>
+    <inputs>
+        <param name="date" type="text" value="yyyy-mm-dd HH:MM" label="date (UTC) as string ('yyyy-mm-dd HH:MM')"/>
```
bgruening, line 32: This is date and time isn't it?

bgruening: What about a special date selector in Galaxy? And a time selector in addition?

tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,91 @@
...
+        <param name="latitude" type="float" min="-90" max="90" value="0"  label="latitude / °N  [-90 90]"/>
+        <param name="longitude" type="float" min="-180" max="180" value="0"  label="longitude / °E [-180 180]"/>
+        <param name="pressure" type="float" value="0"  label="pressure / dbar"/>
+        <param name="temperature" type="float" value="0"  label="in-situ temperature / °C"/>
```
bgruening, line 36: We can add min here and probably also a max?

tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,91 @@
...
+        <param name="variables" type="select" label="Variable names to be predicted" multiple="true">
+                <option value="AT">AT</option>
+                <option value="CT">CT</option>
+                <option value="pH">pH</option>
+                <option value="pCO2">pCO2</option>
+                <option value="NO3">NO3</option>
+                <option value="PO4">PO4</option>
+                <option value="SiOH4">SiOH4</option>
+        </param>
+        <param name="epressure" type="float" value="0.5" label="Default value for input errors pressure (optional) 0.5 dbar" optional="true"/>
```
bgruening, line 48: What is dbar. Can we add max min?

bgruening (commented):
Review incoming!

tools/ocean_neural_network/canyonb.xml
Marie59: This something I remembered you ask me to add in my tools a while back. But I don't remember what literal means I just remember it's to have a trace of the files needed for the tool to work.

tools/ocean_neural_network/canyonb.xml
Marie59: If I remember my physic classes dbar (decibar) = 0,1 bar the pressure unit ;) I am pretty sure we can put min and max as it (same if I remember my classes well) is the acceptable error. But I'll have to ask my coworker the exact min and max we can put

tools/ocean_neural_network/canyonb.xml
bgruening: It does not exist yet, but we could create one for next year if you think it could be useful for many tools and many commuities.

tools/ocean_neural_network/canyonb.xml
bgruening: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-required-files-include

tpye="path" isn't it?

bgruening: Is the galaxy language server complaining?

tools/ocean_neural_network/canyonb.xml
Marie59: Aaah yes it would be very useful. I think almost one out of 2 of the tools I added this last 2 years needed a date.

tools/ocean_neural_network/canyonb.xml
Marie59: all good

Marie59: `.. ERROR (XSD): Invalid XML: Element 'include', attribute 'type': [facet 'enumeration'] The value 'path' is not an element of the set {'literal', 'prefix', 'glob', 'regex'}.`

Marie59:
okay so everything is working here. Testing out the rest

tools/ocean_neural_network/canyonb.xml
bgruening: I only know https://github.com/galaxyproject/tools-iuc/blob/4dba8c02f5b33b425d4ecd9c9f0d0eaa0a74f9bf/tools/extract_genomic_dna/extract_genomic_dna.xml#L8 to be honest.

But you are more correct here. TIL. Thanks!

tools/ocean_neural_network/canyonb.xml
bgruening: Ok, can you please add a comment here. The more detailed the better. E.g. sketching out use-cases. How the UI should look like, but also how the "string" should look like (could be multiple ones) that you get on the command section.

bgruening (commented):
LGTM, let me know when you are finished.

tools/ocean_neural_network/canyonb.xml
Marie59: Okay so the date should be implemented with a kind of choice I think here are the 3 possibilities:
- yyyy
- yyyy-mm
- yyyy-mm-dd

Then with that there could be the option of the hours that would follow the date in the same string ? 
- HH:MM
- HH:MM:SS

There could be 3 possibilities to fill this in:
- We put it as it a string line where the user must replace the "yyyy" and "mm" by integers 
- We put it as a drop box like first you select the year, then the possibility to select the months opens, then the days and so on (I think this the one I like most)
- Or last one you put the date and it opens like a small calendar where you can select year (and) month and so on (like on most website ui) personally this is not the solution i prefer ...

Is this good ? or do you need more detail ?

Marie59:
Workig on putting the test files in the right spot like for the other PR I'm working on

tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,148 @@
...
+            <param name="input_data" location="https://gitlab.com/pokapok-projects/easy-qcv-bgc/libraries/tool_canyonb_galaxy/-/blob/c26f464d79cea10e65c73c83b018841699c732e5/embedded-resources/example-data/full_odv_collection_galaxy_exemple_simplified.csv"/>
+            <param name="type" value="no"/>
+            <output name="output">
+                <assert_contents>
+                    <has_n_lines n="274026"/>
+                </assert_contents>
+            </output>
+        </test>
+        <test expect_num_outputs="1">
+            <param name="input_data" location="https://gitlab.com/pokapok-projects/easy-qcv-bgc/libraries/tool_canyonb_galaxy/-/blob/c26f464d79cea10e65c73c83b018841699c732e5/embedded-resources/example-data/data_from_GEOTRACES_IDP2021_Seawater_Discrete_Sample_Data_v2.txt"/>
```
bgruening, line 72:
```suggestion
            <param name="input_data" location="https://gitlab.com/pokapok-projects/easy-qcv-bgc/libraries/tool_canyonb_galaxy/-/raw/c26f464d79cea10e65c73c83b018841699c732e5/embedded-resources/example-data/data_from_GEOTRACES_IDP2021_Seawater_Discrete_Sample_Data_v2.txt"/>
```

bgruening: that one?

bgruening:
> .. WARNING (HelpInvalidRST): Invalid reStructuredText found in help - [<string>:20: (ERROR/3) Unexpected indentation.

bgruening:
> +Linting tool /home/runner/work/tools-ecology/tools-ecology/tools/ocean_neural_network/canyonb.xml
Failed linting
.. WARNING (TestsCaseValidation): Test 1: failed to validate test parameters against inputs - tests won't run on a modern Galaxy tool profile version. Validation errors are [Invalid parameter name found type]
.. WARNING (TestsCaseValidation): Test 2: failed to validate test parameters against inputs - tests won't run on a modern Galaxy tool profile version. Validation errors are [Invalid parameter name found type]

Marie59:
Here also ?? can I say hurray ??

bgruening:
This will now happen with most of your tools :)
All tests need to be updated over time.

For now, you can say, Hurray!

bgruening:
If you think this is ready, please remove the draft status.

tools/ocean_neural_network/canyonb.xml
bgruening:
```suggestion
```

tools/ocean_neural_network/canyonb.xml
```diff
@@ -0,0 +1,151 @@
...
+            <param name="type" type="select" label="Do you have your own mapping to make ?" help="The mapping is needed for ODV collection extracted from the ODV software.">      
+                <option value="no">No, the mapping is already correct in my data</option>
+                <option value="yes">Yes, I have to make a mapping</option>
+            </param>
+            <when value="no">
+            </when>
+            <when value="yes">
+                <param name="mapping_pres" type="text" label="Column name of the pressure"/>
+                <param name="mapping_temp" type="text" label="Column name of the in-situ temperature"/>
+                <param name="mapping_psal" type="text" label="Column name of the salinity"/>
```
bgruening, line 49: those are columns from the `input_data`?

If so, the `input_data` needs to be CSV not txt, isn't it?

bgruening: can this be a column_type parameter?

Marie59: nope nope this is something that the user needs to fill in manually according to what his data looks like. I put csv or txt format but it can also be odv format (that galaxy reads as a txt) (maybe one day i'll need to really add this format in the galaxy datatype).

Thus, galaxy will no necessarily register that as a column. If we have time next week maybe we can check it out

bgruening (approved):

PR #151, opened 2025-01-16
Update geo_cov_template.xml to remove conda requirement

yvanlebras:
To only have a container requirement. It seems to me mixing both make the tool not working...


Marie59 (approved):
I think it's good !

bgruening (approved):

PR #175, opened 2025-06-12
Imporve extension management

Marie59:
The tools are extension sensitive (netcdf ones) In order to better chain them I improved the management of the extension


Marie59:
@bgruening @yvanlebras ready when you are

Marie59:
not ready wait !!!!!

bgruening:
I'm also not sure this is the correct fix? Can you explain what the problem is that you want to fix?

PR #176, opened 2025-06-27
small update for friendlier use

Marie59:


Marie59:
I cannot test it locally I don't have my good computer with me...
everything seems to run well here from the looks of it

PR #181, opened 2025-08-13
Add tool to import OPeNDAP datasets using xarray

knl88:
This uses xarray and netcdf4 to open remote datasets available over OPeNDAP, I haven't seen a tool for this and think it would fit nicely with the already existing xarray tools.

### DIFF
--- a/tools/xarray_import_data/xarray_import_data.xml
+++ b/tools/xarray_import_data/xarray_import_data.xml
@@ -13,7 +13,9 @@
     <inputs>
         <param name="opendap_url" type="text" format="netcdf" label="OPeNDAP URL"
             help="OPeNDAP URL to a dataset, also see https://docs.xarray.dev/en/stable/user-guide/io.html#opendap"
-            optional="false" />
+            optional="false">
+            <validator type="regex" message="Type a valid url">^(https:|http:|\.)\S*</validator>
+        </param>
         <param name="decode_times" type="boolean" label="Decode times"
             help="If true, decode time variables to datetime objects. If false, keep them as raw data."
             checked="true" />
--- a/tools/xarray_import_data/xarray_import_data.xml
+++ b/tools/xarray_import_data/xarray_import_data.xml
@@ -29,7 +29,7 @@
         <test>
             <param name="opendap_url"
                 value="http://test.opendap.org:8080/opendap/catalog/data/nc/data.nc" />
-            <output name="output_dataset">
+            <output name="output_dataset" ftype="netcdf">
                 <assert_contents>
                     <has_size min="17K" max="18K" />
                 </assert_contents>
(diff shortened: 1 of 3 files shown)


tools/xarray_import_data/xarray_import_data.xml
```diff
@@ -0,0 +1,64 @@
...
+    <requirements>
+        <requirement type="package" version="2025.4.0">xarray</requirement>
+        <requirement type="package" version="3">python</requirement>
+        <requirement type="package" version="1.6.0">netcdf4</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        python '$__tool_directory__/xarray_import_data.py' '$opendap_url' '$decode_times' '$decode_cf' '$output_dataset' 
+    ]]></command>
+    <inputs>
+        <param name="opendap_url" type="text" format="netcdf" label="OPeNDAP URL"
```
bgruening, line 14: what happens if a user puts a github/youtube whatever URL in here?

tools/xarray_import_data/xarray_import_data.xml
```diff
@@ -0,0 +1,64 @@
...
+    </inputs>
+    <outputs>
+        <data name="output_dataset" from_work_dir="xarray_imported_data" format="netcdf"
+            label="xarray dataset" />
+    </outputs>
+    <tests>
+        <test>
+            <param name="opendap_url"
+                value="http://test.opendap.org:8080/opendap/catalog/data/nc/data.nc" />
+            <output name="output_dataset">
```
bgruening, line 32:
```suggestion
            <output name="output_dataset" ftype="netcdf">
```

bgruening (commented):
Can you please fix the Python linting errors? Thanks.

tools/xarray_import_data/xarray_import_data.xml
knl88: Btw, should be easy to add support for s3 URLs ++ if needed.

bgruening: > Btw, should be easy to add support for s3 URLs ++ if needed.

I guess at this stage we should really explore the deferred-dataset use-case a bit more. If you like tell me where you got stuck in testing it.

knl88: Yep, that is a good idea! When I tested deferred dataset it didn't pass the URL, but just an empty string, it was also a bit fiddly to get to the deferred option, but it would be nice having it easily working 👍

bgruening: Here is the schema for the validators: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-validator

knl88: Thanks, yep added a small one, but the url can take quite many forms.

PR #183, opened 2025-08-15
add five new aquaInfra functions

MarkusKonk:


tools/aquainfra_ogc_api_processes/macros.xml
```diff
@@ -41,7 +41,7 @@
                     <param name="input_data" value="points_att_polygon_test_input_3.txt"/>
                     <param name="colname_long" value="longitude"/>
                     <param name="colname_lat" value="latitude"/>
-                    <param name="colname_value_name" value="transparency_m"/>
+                    <param name="colname_value_name" value="transparen"/>
```
bgruening, line 44: mh, this looks wired, can you double check?

MarkusKonk: That's indeed correct since the column in the corresponding data is called "transparen" (for whatever reason :) )

MarkusKonk:
> Kannst du mit die Connection mit GBIF erklären? Interessiert mich :)

I will ask ;).

PR #186, opened 2025-08-28
Qcv feedback

Marie59:
First feedbacks from the users of what needs improvements


tools/ocean_data_view_manager/tool_odv.xml
```diff
@@ -105,7 +105,7 @@
         </conditional>
     </inputs>
     <outputs>
-        <data name="output" format="txt" label="ODV collection manager output"/>
+        <data name="output" format="txt" label="odv_collection.txt"/>
```
bgruening, line 108: This does not look good. Maybe as help=""

PR #187, opened 2025-09-18
add data source tool for NFDI4Earth's OneStop4All

MarkusKonk:
This data source tool redirects users to the search UI of NFDI4Earth's [OneStop4All](https://onestop4all.nfdi4earth.de/). In the future, users will be able to select a dataset in the [search UI](https://onestop4all.nfdi4earth.de/search?sort=) and import it to the Galaxy platform for further analysis via a dedicated "Import to Galaxy" - Button (similar to what we have done in AquaINFRA). We are still working on this but already would like to make it possible for users to get from Galaxy to the OS4All.

### DIFF
--- a/tools/nfdi4earth_os4a_importer/.shed.yml
+++ b/tools/nfdi4earth_os4a_importer/.shed.yml
@@ -1,0 +1,11 @@
+categories:
+    - Ecology
+owner: ecology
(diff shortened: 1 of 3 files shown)


wm75 (commented):
Wondering how many different data formats users might end up downloading with this extremely generic tool, but I guess you can refine things later.

MarkusKonk:
It's basically the same as in this previous PR but for another platform: https://github.com/galaxyecology/tools-ecology/pull/108

PR #200, opened 2025-11-21
add new aquainfra tool: Copernicus CDS Downloader for AquaINFRA marine model

MarkusKonk:


tools/ccds_download_era5_af/get_era5.py
```diff
@@ -0,0 +1,64 @@
...
+
+start_date = datetime.strptime(SDATE, "%Y-%m-%d")
+end_date = datetime.strptime(EDATE, "%Y-%m-%d")
+
+date_list = list(
+    pd.date_range(start_date, end_date, freq='D')
+    .strftime('%Y-%m-%d')
+)
+
+VARIABLES = ['10m_u_component_of_wind',
```
bgruening: you can remove this entire file and avoid passing around params by using something like this:

https://github.com/galaxyproject/tools-iuc/blob/eceb4b3a69c5a755feac55801e85eb914a669042/tools/query_tabular/filter_tabular.xml#L16

You probably have no time to do this now, just wanted to point that out.

For small scripts, this is very handy imho

PR #209, opened 2025-12-02
Restore copernicusmarine.xml tool due to version mismatch 

TuturBaba:


TuturBaba:
I would like to run **Copernicus Marine Data Store**, but I’m getting the following error: ``` ERROR - 2025-11-28T09:07:49Z - Client version 1.3.3 is not compatible with current backend service. Please update to the latest client version.```

I updating the version, but I’m unable to test it locally on my PC, the process just hangs indefinitely.

TuturBaba:
@yvanlebras @bgruening I don’t understand why the test passes on GitHub, but when I check the log, I see “copernicusmarine (Test #1): failed.”

PR #212, opened 2026-01-06
Wildlife: Error NMS

TuturBaba:
I have an issue with the Wildlife tool. It works correctly on my computer when running on CPU, but in Galaxy with GPU I get the error: "NMS time limit 0.850s exceeded".
As I understand it, the problem is that the GPU is faster and tries to process more images, which increases the processing time of the NMS step. To handle this, the system removes or skips some detections to avoid exceeding the time limit.
Therefore, one possible solution would be to reduce the batch_size.


bgruening:
@TuturBaba can you please try again on EU. I took over your changes, so you can test it on GPU.

PR #214, opened 2026-01-19
medenv: fix Copernicus credentials accessibility 

PaulineSGN:

PR #215, opened 2026-01-21
CMEMS downloader: switch to new crendentials method in 25.1

MarkusKonk:


bgruening:
Can you please also look at https://github.com/galaxyecology/tools-ecology/pull/214/changes and see if you want to align with the variable names etc for the same service?

MarkusKonk:
> So I tested the tool using Planemo and found out that the bc requriement is causing issues here. When I remove it, it works using Planemo and when I add it, I get the netCDF4 not found issue mentioned above. I am not sure if it's really needed on the Galaxy server. I think we added it for some reason...Shall we try it without bc?

Indeed the test fails without bc
image

PR #231, opened 2026-03-06
Update copernicusmarine.xml to relaunch CI and try push on toolshed

yvanlebras:


yvanlebras:
Was testing if we can update the tool on TS with the updated ci but seems this is not working... Maybe I need to bump the tool version to have lint ok? In the meantime we can normally apply simple update without bumping version isn't it ? Additionally, it seems the ci wz'as ok on 3th of march as medenv was updated on TS at this date... Any idea @PaulineSGN  @bgruening ?

PR #243, opened 2026-03-23
Sam3 Add parametre Normalize

TuturBaba:


yvanlebras (approved):

yvanlebras:
ok ;) the error is:
```
/home/runner/work/_actions/peter-evans/create-or-update-comment/v5/dist/index.js:2436
        throw new Error('Parameter token or opts.auth is required')
```
I need to look at the CI files

yvanlebras:
I checked that IUC CI/CD is using `peter-evans/create-or-update-comment/v5` and it is the case... so the origin of the error seems to not be related to an update of this... So it appears to me the error is related to credentials

yvanlebras:
Oh, in fact this is ok! https://toolshed.g2.bx.psu.edu/view/ecology/sam3_semantic_segmentation/0bc5b1bfa81a

PR #250, opened 2026-04-03
Sam3 add option (normalize and quality)

TuturBaba:


yvanlebras (approved):
THANK  You Arthur ! Seems to me ok ! This will enhance ++ the useability of the tool!

bgruening (commented):
Nice, very cool. Can you please add a test for the new functionality?

TuturBaba:
> Nice, very cool. Can you please add a test for the new functionality?

I'm not sure what kind of test you want. The only existing test checks for a failure case: `<test expect_exit_code="1" expect_failure="true">`
I can add a test that uses the new functionality, new inputs but it will still test a failure case. Is that what you mean?

bgruening:
A test that is testing the different output filters for example. That we know the tool logic is working?

TuturBaba:
> A test that is testing the different output filters for example. That we know the tool logic is working?

I'm sorry, I'm not sure I understand what you mean. I can add this test  
```        
<test expect_exit_code="1" expect_failure="true">
    <param name="sam3_models" value="unknown"/>
    <conditional name="input">
        <param name="input_kind" value="video" />
        <param name="source" value="video-elephant.mp4"/>
        <param name="quality" value="copy"/>
        <param name="coco_video_mode" value="video"/>
        <param name="outputs_format" value="yolo_bbox"/>
    </conditional>
    <param name="text_prompt" value="elephant"/>
    <param name="conf" value="0.25"/>
    <param name="vid_stride" value="5"/>
    <assert_stdout>
        <has_text text="Invalid model!"/>
    </assert_stdout>
</test>
``` 
It has the same failure case as the other one, but with different inputs (new feature). However, I can't test a wrong input for the new options, because they are all select inputs, so Galaxy will not even run the test if the value is invalid.

bgruening:
Ok, thanks for trying!

yvanlebras:
THANK YOU Arthur, Björn ! Have a nice end of week, and week-end !

PR #257, opened 2026-05-13
Sam3: new help, new parameter (imgsz)

TuturBaba:
Add imgsz parameter (supported inference image sizes: 512, 640, 1024, 1280) passed to the SAM3 predictor to reduce out-of-memory issues during inference
Add validators for image and video input extensions
Add in Help section with troubleshooting guidance for missing or invalid file extensions

### DIFF
--- a/tools/Sam3/sam3_semantic_segmentation.xml
+++ b/tools/Sam3/sam3_semantic_segmentation.xml
@@ -1,4 +1,4 @@
-<tool id="sam3_semantic_segmentation" name="SAM3 Semantic Segmentation" version="1.0.1+galaxy5" profile="25.1">
+<tool id="sam3_semantic_segmentation" name="SAM3 Semantic Segmentation" version="1.0.1+galaxy4" profile="25.1">
     <description>
         SAM3 performs text-prompted semantic segmentation on images or videos. 
     </description>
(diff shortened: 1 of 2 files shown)


tools/Sam3/sam3_semantic_segmentation.xml
```diff
@@ -59,6 +60,7 @@
             </param>
             <when value="image">
                 <param name="source" type="data" format="jpg,png,tiff" multiple="true" label="Input images">
+                    <validator type="expression" message="Your image file must have a valid extension (e.g. rename 'image' to 'image.png'). Supported formats: .jpg, .png, .tif">value.name.lower().endswith(('.jpg', '.jpeg', '.png', '.tiff', '.tif'))</validator>
```
bgruening: Why is this mandatory, and why can we not do this in the cammand section with the link-trick?

`ln -s '$input' foo.{$input.ext}`.

TuturBaba: Yes, the SAM3 function needs the file extension as input. 
I just didn't think of that, thanks!

yvanlebras (approved):
Looks good to me, THANK YOU Arthur, THANK YOU Björn!

PR #262, opened 2026-05-26
Update CRSconverter.xml to add EDAM topics

yvanlebras:


bgruening (commented):
Can we not use macros here?

bgruening:
Should we merge and not deploy? Wait for regular updates and then with the next update the edam will be added? Or do you want to deploy all of the tools? Test would need to pass and maybe we need to bump the tool versions.

PaulineSGN:
> Should we merge and not deploy? Wait for regular updates and then with the next update the edam will be added? Or do you want to deploy all of the tools? Test would need to pass and maybe we need to bump the tool versions.

Either option works for us. If we merge and deploy, the tests need to pass — but as I understand it, they won't right now because the EDAM topics we want aren't in the current Galaxy version. Is there a place where we could add those EDAM terms ourselves before the next Galaxy update, so we could deploy now instead of waiting for it? Or am I missing something?

ping @yvanlebras

bgruening:
Mh why are those to terms missing? Galaxy has the latest EDAM version.

PaulineSGN:
> Mh why are those to terms missing? Galaxy has the latest EDAM version.

I don't know, that's what the lint failure says, for example:

`.. WARNING (EDAMTermsValid): No entry '
            topic_0610
        ' in EDAM.
.. WARNING (EDAMTermsValid): No entry '
            topic_3050
        ' in EDAM.
`

PR #266, opened 2026-06-10
Update RLDNe tool 

GINAMO-EBVs:
Add a new option : correction Nb to Ne


yvanlebras:
Thank you Laura ! Don't hesitate to say when review need to be done. Don't hesitate to use the option "draft Pull Request" to add you commits in a ""on going mode"" then, when ready, to convert the "draft PR" to a "normal PR".

GINAMO-EBVs:
> Thank you Laura ! Don't hesitate to say when review need to be done. Don't hesitate to use the option "draft Pull Request" to add you commits in a ""on going mode"" then, when ready, to convert the "draft PR" to a "normal PR".

Hi @yvanlebras , it’s ready for review! I’ve had the issue with the missing file error message again, so I’ve changed the file extension from ‘.r’ to ‘.R’ in the hope that it will work.

tools/RLDNe/RLDNe.xml
```diff
@@ -92,6 +105,45 @@
...
+                        help="Coefficient of variation of bx (mean number of offspring produced at age x) for adult lifespan (using only ages with bx>0)."/>
+
+                    <param name="variation_coefficient_bx" type="float" optional="true"
+                        label="Coefficient of variation of breeding"
+                        help="Required if the above option is checked. Coefficient of variation of bx (mean number of offspring produced at age x) for adult lifespan (using only ages with bx>0)."/>
+                </when>
+
+            </conditional>
+        </section>
+
```
PaulineSGN, line 146: Hi! Thank you for your work! 
Here, if I understand correctly, when the cohort is a mixed cohort, the tool doesn't do anything about it? And then, when the user selects "single cohort", there is some treatment on the data. So here, maybe we just need a boolean to know if the cohort is a single cohort or not instead of a select? What do you think about it?

tools/RLDNe/RLDNe.xml
```diff
@@ -269,6 +380,36 @@ This tool implements the Linkage Disequilibrium (LD) method from **NeEstimator V
...
+.. math::
+
+    \frac{N_b}{N_e} = 0.833 + 0.637 \times \log(AL) - 0.423 \times C_{Vf}
+
+(Waples *et al.* 2013, Figure 3b)
+
+The resulting adjusted value is appended to the output table under the column **NeLD_with_correct_pseudorep_Nb**.
+
+----
+
```
PaulineSGN: Maybe we can swap the "Correction : From effective number of breeders (Nb) to effective population size (Ne)" section and the "Pseudoreplication correction" section to have them in the same order as the tool parameters?

tools/RLDNe/RLDNe.xml
GINAMO-EBVs: Hi @PaulineSGN, yes exactly ! I used 'select' instead of a boolean because otherwise the conditional wouldn't work : 'WARNING (ConditionalParamTypeBool): Conditional [mixed_cohort_params] first param of type="boolean" is discouraged, use a select'. Otherwise, I'd have to remove the conditional and make the other paramters optional.

PaulineSGN: Oh yes, you're right, sorry! So maybe you can use `display= radio'` but it's just an aesthetic thing if you want.

PaulineSGN (approved):

tools/RLDNe/RLDNe.xml
GINAMO-EBVs: Good idea, I'll change that! Thanks!

PR #280, opened 2026-07-07
Add pivot wider tool

PaulineSGN:


tools/Ecoregionalization_workflow/pivot_wider.xml
```diff
@@ -0,0 +1,181 @@
+<tool id="pivot_wider_ecoregion" name="Pivot Wider" version="0.1.0+galaxy0" profile="25.1">
+    <description>
+        for ecoregionalization 
+  </description>
+    <requirements>
+        <requirement type="package" version="4.5">r</requirement>
+        <requirement type="package" version="1.3.2">r-tidyr</requirement>
+    </requirements>
```
bgruening, line 8: please add required_files

tools/Ecoregionalization_workflow/pivot_wider.xml
```diff
@@ -0,0 +1,181 @@
...
+            '$names_from'
+            '$values_from'
+            '$values_fill'
+            '$values_fn'
+            '$replace_occ'
+   ]]>
+    </command>
+    <inputs>
+        <param name="table" type="data" format="tabular" label="Input your table here please" help="See example below"/>
+        <param argument="--names_from" type="data_column" data_ref="table" label="Which column (or columns) to get the name of the output column from" multiple="false"/>
```
bgruening: multiple=false, but the text states that multiple can be submitted

bgruening (commented):
Imho this tools is too narrowed. In my opinion we should aim for more generalized tools, maybe like https://github.com/bgruening/galaxytools/blob/master/tools/statistics/statistical_hypothesis_testing.xml

tools/Ecoregionalization_workflow/pivot_wider.xml
PaulineSGN: Oops, sorry, last-minute changes; I'll change that.

PaulineSGN:
> Imho this tools is too narrowed. In my opinion we should aim for more generalized tools, maybe like https://github.com/bgruening/galaxytools/blob/master/tools/statistics/statistical_hypothesis_testing.xml

You're right, I've generalized it as much as I could, but this is just a first step that responds to our immediate needs. It might be enhanced further later on, notably by implementing other features of the pivot_wider function.

PR #281, opened 2026-07-08
Get era5 data

Mathmath34:
This tool allows one to download ERA5 data from the Climate Data Store of Copernicus.


Marie59:
Hmmm not sure this is usefull check this tool out https://earth-system.usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu/repos/climate/c3s/c3s/0.3.1

Mathmath34:
This tool (https://earth-system.usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu/repos/climate/c3s/c3s/0.3.1) is probably similar to the one I created but it also has some differences:
-From what I understood C3S is a generic pass-through to the raw CDS API: it requires a CDS account, the request has to be hand-built on the CDS website and pasted in, and it returns gridded NetCDF/GRIB (hourly, or monthly means).

-uses the same underlying ERA5/ERA5-Land reanalysis (redistributed by the Open-Meteo API), but needs no account, takes a simple coordinate file + variable picklist, and outputs a ready-to-use daily table (CSV/TSV/JSON/Parquet) per site.

So, even though, it is the same data, the access path is different. I don't know if that makes the tool relevant for publication though.

Marie59:
Hmm I am wondering ... would it make sense then to have 1 tool and you can choose / have the option I you want to use it ? or 2 different tools would be better ? 
any advise here @bgruening @yvanlebras @PaulineSGN ?

tools/get_ERA5_data/get_ERA5_data_openmeteo.xml
```diff
@@ -0,0 +1,402 @@
+<tool id="era5_extractor" name="ERA5 / ERA5-Land Climate Data Extractor" version="3.0.0" profile="25.1">
+
+  <description>Download daily ERA5 / ERA5-Land climate data via Open-Meteo (no account required)</description>
```
bgruening:
```suggestion
  <description>Download daily ERA5 / ERA5-Land climate data via Open-Meteo</description>
```

tools/get_ERA5_data/get_ERA5_data_openmeteo.xml
```diff
@@ -0,0 +1,402 @@
...
+             applies its own internal validator which conflicts with regex
+             validators in Galaxy profile 23.0.  Date format validation is
+             handled by the Python script itself. -->
+        <param name="start_date"
+               type="text"
+               value="1990-01-01"
+               label="Start date"
+               help="Format: YYYY-MM-DD. Earliest allowed: 1960-01-01 (SAFRAN lower bound).
+                     Overridden per row by a 'begin_date' column in the coordinate file." />
+        <param name="end_date"
```
bgruening, line 148: Here a nice validator would be beneficial

tools/get_ERA5_data/get_ERA5_data_openmeteo.xml
```diff
@@ -0,0 +1,402 @@
...
+             type="select"
+             label="Date range source"
+             help="Global: same start/end dates applied to every site.
+                   Per-site: dates are read from 'begin_date' and 'end_date'
+                   columns in the coordinate file (no global dates needed).">
+        <option value="global" selected="true">Global dates — same period for all sites</option>
+        <option value="file">Per-site dates — from coordinate file columns</option>
+      </param>
+      <when value="global">
+        <!-- FIX: no <validator> on text date params — planemo 3 / pydantic v2
```
bgruening: Mh, this looks wrong. Can you add a validator and show us the error that you get?

tools/get_ERA5_data/get_ERA5_data_openmeteo.xml
```diff
@@ -0,0 +1,402 @@
...
+             value="1.0"
+             min="0.0"
+             max="60.0"
+             label="Sleep between API calls (seconds)"
+             help="Minimum pause between successive HTTP calls (default: 1.0 s).
+                   The free tier allows at most 600 requests/min, 5000/hour, 10000/day.
+                   Increase this value if you receive repeated HTTP 429 errors.
+                   Set to 0 only when using the paid API key." />
+
+      <param name="verbose"
```
bgruening: is this only for admins?, if so make it a type="hidden" param or remove it.

bgruening (commented):
Thanks a lot! Small comments inline, please also add a `<required_files>`.

@Marie59 which two tools would that be?

Marie59:
> Thanks a lot! Small comments inline, please also add a `<required_files>`.
> 
> @Marie59 which two tools would that be?

This tool from @annefou https://usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu/repos/climate/c3s/c3s/0.3.1 which allow to retrieve data from copernicus climate like here (such as ERA5 data). But it needs a user account and to copy paste the API commandline from the website.

@Mathmath34's tool on the other hand online retrieve ERA5 data from Copernicus climate and does not need any user account or to go on the website 

So should this be 2 seperated tools ? or should we do an update of the 1st one to allow different retrieving data possibilities ??

PR #286, opened 2026-08-13
add required files, creator and profiles

Marie59:
First folder 1 still some to go ! A bit of cleanups/ updates on the tools to add all the right metadatas, will do the same on earth side

Useless stuffs ?:
- [ ] Is this still necessary https://github.com/Marie59/tools-ecology/blob/update-metadata/tools/ccds_download_era5_af/get_era5.py
- [ ] Is this still necessary https://github.com/Marie59/tools-ecology/blob/update-metadata/tools/cmems_download_ic_bc/medsea_bc_limits.py
- [ ] Is this necessary https://github.com/Marie59/tools-ecology/blob/update-metadata/tools/stoc/test.bash.out and https://github.com/Marie59/tools-ecology/blob/update-metadata/tools/stoc/test.bash

CREATORS
- [x] @PaulineSGN  add orcid if you wish instead of git 
- [x] @C0c0-C00k13 Do you want to add your family name in the creator ?
- [ ] https://github.com/Marie59/tools-ecology/blob/update-metadata/tools/retrieve_bold/retrieve_bold.xml => check if I put the right name for @triskell-cumunel 
- [x] @Morphy427 CRSConverter EstimEndem PhylOccuMatcher PhyloIndex => I put PNDB as creator I don't know who is Morphy
- [ ] @GINAMO-EBVs RLDNe SSR_filtering genetic_clustering genetic_format_conversion population_genetics_stats vcftoolbox => I put PNDB as creator I don't know who is Ginamo
- [x] @Rassine sdmpredictors => I put PNDB as creator I don't know who is Rassine

Test

- [ ] tools/Geom_mean_workflow/Moyenne_geom.xml  <param name='separator' value="&#44;"/>
- [ ] tools/regionalgam/flight-curve.xml ?
- [x] tools/retrieve_bold/retrieve_bold.xml ?

PR #288, opened 2026-08-19
create deprecated tools

Marie59:
open new PR for the deprecated tools found here https://github.com/galaxyecology/tools-ecology/pull/286


Marie59:
@bgruening this is ready too I think just it never finishes to run

PR #291, opened 2026-08-25
Fix deprecated

Marie59:


Marie59:
@bgruening could you merge that if you are okay with it ? (Maybe it's not enough it fixes the ci run)

bgruening (approved):

PR #292, opened 2026-08-28
add interactive folder

Marie59:
Move around interactive folder


yvanlebras (approved):

yvanlebras:
Very good idea THANK YOU Marie !

bgruening (approved):
But I would actually remove them. I don't think they are up to date and in sync with the deployed ones.

yvanlebras:
After dscussing with Marie and Pauline, it is a good idea waiting to have a proper manner to manage GxIT exactely as classical tools in the Toolshed and ci and so on! THANK YOU !!

PR #294, opened 2026-09-08
Minor Sam3 paper cuts 

bernt-matthias:


Marie59 (approved):
Thanks @bernt-matthias !

bernt-matthias:
Was wondering if people really select tracks manually or rather filter by pvalue?

yvanlebras:
THANK YOU Matthias, Marie. Seems to me it is better with your commits Matthias ! Considering your comment, you are speaking with frames selection ( https://github.com/bernt-matthias/tools-ecology/blob/sam3-fixes/tools/Sam3/sam3_semantic_segmentation.xml#L84C30-L84C43 ) ?

bernt-matthias:
The output text annotations seem to contain probability values. I was wondering if it would be a good idea to use this first for filtering?

